PR #2, opened 2015-12-08
Fix to infinite recursion in data/timeseries calls

### LINKED ISSUE #1: Calls to timeseries/data without a polygon raise max recursion depth
### To reproduce

Start a dev server with real world data using the production mddb
`MDDB_DSN=postgresql://httpd_meta@monsoon.pcic.uvic.ca/pcic_meta python scripts/devserver.py -p 8004`

#### Timeseries

Request
`http://<root_url>/api/timeseries?id_=tmax_monClim_PRISM_historical_run1_198101-201012&variable=tmax&area=`
Response
`RuntimeError: maximum recursion depth exceeded in __instancecheck__`
[Full stack trace](http://hastebin.com/ufucoxigid.coffee)

#### Data

Request
`http://atlas.pcic.uvic.ca:8004/api/data?model=PRISM&emission=historical&time=0&variable=tmax&area=`
Response
`RuntimeError: maximum recursion depth exceeded while calling a Python object`
[Full stack trace](http://hastebin.com/loxevafixi.coffee)

jameshiebert:
Fixes #1. Reading an array from a NetCDF file can either return a `numpy.ma.MaskedArray` (if some data are masked out, such as over the ocean) or a plain `numpy.array`. The previous call to `get_array()` (and all of the tests) assumed a plain array.

This bugfix simply adds a check to see if we have a mask, and if so it updates the existing mask with the polygon mask.


jameshiebert:
It should be good to go, but feel free to review.

PR #5, opened 2015-12-17
Wrote the multistats API call to return stats from multiple things

### LINKED ISSUE #4: Write a `multistats` API call
Takes a model_id and emissions scenario, searches for unique_ids and delegates to the stats() API call.

jameshiebert:
Gives the ability to request stats for multiple models and scenarios in a single request

Closes #4

### DIFF
--- a/ce/api/stats.py
+++ b/ce/api/stats.py
@@ -9,6 +9,11 @@ from modelmeta import DataFile, Time
 
 from ce.api.util import get_array, get_units_from_netcdf_file, mean_datetime
 
+na_array_stats = {
+    key: np.nan
+    for key in ('min', 'max', 'mean', 'median', 'stdev', 'ncells')
+}
+
 def stats(sesh, id_, time, area, variable):
     '''Request and calculate summary statistics averaged across a region
 
@@ -50,11 +55,18 @@ def stats(sesh, id_, time, area, variable):
                 }
             }
 
-    Raises:
-        Exception in several cases:
+        There are two cases semi-error cases which should be mentioned,
+        when the filesystem is out of sync with the database.
 
         1. The file pointed to by `id_` does not exist in the filesystem
         2. The requested variable does not exist in the given file
+
+        In these cases, the numerical values will all be NaN, and the
+        results dict will be missing the 'units' and 'time' keys.
+
+    Raises:
+        None?
+
     '''
     try:
         df = sesh.query(DataFile).filter(DataFile.unique_id == id_).one()
@@ -62,7 +74,10 @@ def stats(sesh, id_, time, area, variable):
     except NoResultFound:
         return {}
 
-    array = get_array(fname, time, area, variable)
+    try:
+        array = get_array(fname, time, area, variable)
+    except:
+        return {id_: na_array_stats}
     stats = array_stats(array)
 
     query = sesh.query(Time.timestep).filter(Time.time_set_id == df.timeset.id)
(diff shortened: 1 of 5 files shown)


ce/api/stats.py
```diff
@@ -50,11 +55,18 @@ def stats(sesh, id_, time, area, variable):
                 }
             }
 
-    Raises:
-        Exception in several cases:
+        There are two cases semi-error cases which should be mentioned,
```
basilveerman: Words.

ce/api/util.py
```diff
@@ -5,6 +5,7 @@
 import numpy.ma as ma
 from netCDF4 import Dataset
 from shapely.wkt import loads
+from modelmeta import *
```
basilveerman: Normally this is what I do... and then you fix my code with named imports :wink:

ce/api/util.py
```diff
@@ -80,3 +81,19 @@ def mean_datetime(datetimes):
     timestamps = [ dt.replace(tzinfo=timezone.utc).timestamp() for dt in datetimes ]
     mean = np.mean(timestamps)
     return datetime.fromtimestamp(mean, tz=timezone.utc)
+
+def search_for_unique_ids(sesh, ensemble_name='ce', model='', emission='', variable='', time=0):
+    query = sesh.query(DataFile.unique_id)\
+                  .join(DataFileVariable, EnsembleDataFileVariables, Ensemble, Run, Model, Emission)\
+                  .filter(Ensemble.name == ensemble_name)\
+                  .filter(DataFileVariable.netcdf_variable_name == variable)\
+                  .filter(Time.time_idx == time)\
```
basilveerman: What's the purpose of filtering on time_idx?

ce/api/util.py
```diff
@@ -80,3 +81,19 @@ def mean_datetime(datetimes):
     timestamps = [ dt.replace(tzinfo=timezone.utc).timestamp() for dt in datetimes ]
     mean = np.mean(timestamps)
     return datetime.fromtimestamp(mean, tz=timezone.utc)
+
+def search_for_unique_ids(sesh, ensemble_name='ce', model='', emission='', variable='', time=0):
+    query = sesh.query(DataFile.unique_id)\
+                  .join(DataFileVariable, EnsembleDataFileVariables, Ensemble, Run, Model, Emission)\
+                  .filter(Ensemble.name == ensemble_name)\
+                  .filter(DataFileVariable.netcdf_variable_name == variable)\
```
basilveerman: Why is `variable` treated any differently than `model`/`emission`?

ce/api/util.py
```diff
@@ -80,3 +81,19 @@ def mean_datetime(datetimes):
     timestamps = [ dt.replace(tzinfo=timezone.utc).timestamp() for dt in datetimes ]
     mean = np.mean(timestamps)
     return datetime.fromtimestamp(mean, tz=timezone.utc)
+
+def search_for_unique_ids(sesh, ensemble_name='ce', model='', emission='', variable='', time=0):
```
basilveerman: Good to split this out.

ce/api/util.py
jameshiebert: Yeah, I don't *normally* like to do this, but we traverse across 7 ORM classes in the search, and that felt too verbose for me.

ce/api/util.py
jameshiebert: If you're searching for seasonal averages (e.g. pseudo-timestep 14), but there exist files that only have monthly averages, (e.g. 0-11), those files should not be included in the search results.

ce/api/util.py
jameshiebert: A single file can potentially contain multiple variables (though it's not recommended), but only ever represents a single model and a single emission scenario.

PR #10, opened 2016-01-12
Feature/memoize geo

basilveerman:
Add memoized wrapper around `polygonToMask`. Memoizing is keyed to the netCDF `model_id` and the WKT polygon

### DIFF
--- a/ce/api/geo.py
+++ b/ce/api/geo.py
@@ -86,7 +86,7 @@ class memoize_mask(object):
         with cache_lock:
             self.cache[key] = result
             self.misses += 1
-            if getsize(self.cache) > self.maxsize * 1024 * 1024: # convert to MB
+            while getsize(self.cache) > self.maxsize * 1024 * 1024: # convert to MB
                 self.cache.popitem(0) # Purge least recently used cache entry
 
         return result
--- a/ce/api/geo.py
+++ b/ce/api/geo.py
@@ -1,5 +1,6 @@
 import logging
 import collections
+import sys
 from collections import OrderedDict
 from threading import RLock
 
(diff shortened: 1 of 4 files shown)


ce/api/geo.py
```diff
@@ -1,6 +1,101 @@
...
+            except KeyError:
+                pass
+
+        log.debug('Cache MISS')
+        result = self.func(*args)
+
+        with cache_lock:
+            self.cache[key] = result
+            self.misses += 1
+            if getsize(self.cache) > self.maxsize * 1024 * 1024: # convert to MB
```
jameshiebert: If the size of the added item is greater than that of the popped item, the size of the cache will still be above the threshold even after popping. There are a number of cases where this could degenerate. You may want to do this in a while loop instead of a single conditional.

ce/api/geo.py
```diff
@@ -1,6 +1,101 @@
...
+        '''
+
+        self.hits = self.misses = 0
+        self.func = func
+        self.maxsize = maxsize
+        self.cache = OrderedDict()
+
+    def __call__(self, *args):
+
+        # Set key to model_id and wkt polygon
```
jameshiebert: This is definitely an big assumption that may or may not hold. In the context of a "Climate Explorer for CMIP5 data" it will, but if we even try to put CMIP5 and CMIP3 data on the app, this will break (well, not break, but return incorrect results). So, we should either qualify this with massive red flags somewhere, or use actual grid definitions for the keys. The latter may be better anyway, since some models share a common grid.

ce/api/geo.py
```diff
@@ -1,6 +1,101 @@
+import logging
+import collections
+import sys
+from collections import OrderedDict
+from threading import RLock
```
jameshiebert: If we use a multiprocess (rather than multithreaded) approach to handling requests, do we even need to lock the cache?

ce/api/geo.py
```diff
@@ -1,6 +1,101 @@
...
+### End from http://stackoverflow.com/a/30316760/597593
+
+log = logging.getLogger(__name__)
+cache_lock = RLock()
+
+class memoize_mask(object):
+    '''
+    Decorator. Caches wktToMask keyed to model_id and the WKT string
+    '''
+    def __init__(self, func, maxsize=50):
```
jameshiebert: How many polygons will we be able to cache w/ 50 MB? Is this sizing appropriate?

basilveerman: In my tests, each addition to the cache added ~9k to the cache so ~5800 requests at 50 MB. Probably too large with current data, but this will be affected by different grid resolutions. I would think enough space to cache each unique_id with 5 polygons would be sufficient.

Initially the max was enforced to 100 requests by default and then later changed to a memory size limit.

ce/api/geo.py
basilveerman: Correct, we would not. Currently, we enable the use of threading through Flask and have not explicitly ruled out multiple threads for a production environment.

ce/api/geo.py
basilveerman: I *think* model_id should actually be sufficient to differentiate GCM models regardless of CMIP3/5/future. Each model will have a unique name.
You're right that this will break when models like PRISM are calculated at different resolutions, or any downscaled product is generated over different spatial extents.

I'm going to leave it this way for now but add this as an issue to address later on.

basilveerman: Opened #11

Issue #8, opened 2015-12-22
Clipping by polygon greatly increases timeseries generation


basilveerman:
Timeseries calls tend to have significantly more slowdown when querying a polygon region

Using this polygon: `POLYGON+((-135.05859375+52.98828125%2C+-131.05468750000003+49.1796875%2C+-125.97656250000001+46.640625%2C+-122.36328125000001+46.640625%2C+-122.4609375+48.7890625%2C+-114.2578125+48.984375%2C+-119.62890625000001+53.37890625%2C+-135.05859375+52.98828125))`

| Endpoint | Time w/out poly | Time w/ Poly | Times slower |
|-------|-----------------------|-------------------|---------------|
| Timeseries | 0m0.093s | 0m9.163s | 98.5 |
| Data | 0m0.105s | 0m1.128s | 10.7 |
| Multistats | 0m0.069s | 0m1.136s | 16.5 |

| Endpoint | Calls to `geo.polygonToMask` |
|-------------|--------------------------------|
| Timeseries | 17 |
| Data | 2 |
| Multistats | 1 |

### Data

Timeseries w/out polygon
```
[...]

PR #12, opened 2016-01-12
Bugfix/polygon performance

basilveerman:
Somewhat solves #8.

| Endpoint | Time w/out poly | Time w/ Poly | Times slower | New Time |
|-------|-----------------------|-------------------|---------------|--------------|
| Timeseries | 0m0.093s | 0m9.163s | 98.5 | **0.302** |
| Data | 0m0.105s | 0m1.128s | 10.7 | **0.325** |
| Multistats | 0m0.069s | 0m1.136s | 16.5 | **0.364** |

Furthermore, changes introduced in #10 reduce times to nearly that of w/out polygon for future requests.

### DIFF
--- a/ce/api/geo.py
+++ b/ce/api/geo.py
@@ -134,10 +134,12 @@ def polygonToMask(nc, poly):
     min_height = np.min(np.diff(nclons))
     min_cell_area = min_width * min_height
 
-    # Buffer the polygon by 1/2 the diagonal distance
-    # FIXME: may increase larger selections which don't need this to contain more grid cells than intended
-    dist = math.sqrt(math.pow(min_width,2)+math.pow(min_height,2))/2
-    poly = poly.buffer(dist)
+    if poly.area < min_cell_area * 2:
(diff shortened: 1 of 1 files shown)


ce/api/geo.py
```diff
@@ -128,19 +129,20 @@ def polygonToMask(nc, poly):
...
     min_cell_area = min_width * min_height
-    if poly.area < min_cell_area:
-        # We can't just naively mask based on grid centroids.
-        # The polygon could actually fall completely between grid centroid
-        minx -= min_width
-        maxx += min_width
-        miny -= min_height
-        maxy += min_height
+
+    if poly.area < min_cell_area * 2:
```
jameshiebert: This is great. A polygon smaller than the minimum cell area is the degenerative and presumably rare case, so moving the expensive buffering into this branch is definitely the way to go.

If and when we decide to add the capabilities of doing the calculation based on area fractions, we'll have to adjust the math. But this is a good first step.

jameshiebert:
Looks great to me. Thanks for the work on this!

PR #14, opened 2016-01-14
Allow 0 as a valid time parameter

basilveerman:
Old boolean check would return data for all time steps when requesting January resulting in something very close to the annual value.
As long as this doesn't break anything, the fix is very simple.


jameshiebert:
Crap... we were still thinking in terms of 1-17 when we wrote that check. Good catch. The fix looks fine to me.

PR #17, opened 2016-01-20
Rewrote mulitmeta API call to only make one DB query.

### LINKED ISSUE #9: `multimeta` API call is too slow for moderate number of files
I have an sqlite database which has indexed 660 ClimDEX files, and the multimeta API call takes 30 seconds to return.

```bash
$ time curl -s http://localhost:9000/api/multimeta
real	0m29.051s
user	0m0.016s
sys	0m0.000s
```

PostgreSQL may be more reasonable, but if not, we'll need to do some query optimization for the multimeta call.

jameshiebert:
Fixes #9 

This change fixes the root of the performance problem for `multimeta` by making a single database query and constructing the response from the resulting rows (instead of making a subsequent query for each file to get the variables).

The one problem with this fix is that it eliminates the "time" key from the metadata dictionaries in the `multimeta` response. We talked about doing this "eventually", but this should not be merged until the front-end has been adapted to work w/o time values up front.

Here are the performance differences running on an sqlite3 database with ~7000 files.

```bash
#On master branch
hiebert@cheetah:~$ time curl -s localhost:8000/api/multimeta > /dev/null

real	8m13.092s
user	0m0.008s
sys	0m0.010s

# On bugfix-9 branch
hiebert@cheetah:~$ time curl -s localhost:8000/api/multimeta > /dev/null

real	0m0.218s
user	0m0.002s
sys	0m0.004s
```

Seems to save about 200-300 ms *per dataset*, which adds up quickly.

### DIFF
--- a/ce/api/multimeta.py
+++ b/ce/api/multimeta.py
@@ -38,10 +38,12 @@ def multimeta(sesh, ensemble_name='ce', model=''):
 
     q = sesh.query(DataFile.unique_id, Model.organization, Model.short_name,
             Model.long_name, Emission.short_name, Run.name,
-            DataFileVariable.netcdf_variable_name, VariableAlias.long_name)\
+            DataFileVariable.netcdf_variable_name, VariableAlias.long_name,
+            Time.time_idx, Time.timestep)\
             .join(Run).join(Model).join(Emission).join(DataFileVariable)\
             .join(EnsembleDataFileVariables).join(Ensemble)\
-            .join(VariableAlias).filter(Ensemble.name == ensemble_name)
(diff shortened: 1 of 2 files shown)


basilveerman:
Looks good. Opened pacificclimate/climate-explorer-frontend#12. This can be merged in once that is complete.

jameshiebert:
If we want to get this merge over with before dealing w/ the front-end side, I've added a patch that puts the time key back into `multimeta`. Then we can just reverse-apply this patch after the front-end is no longer dependent on the time metadata.

basilveerman:
I'm working on pacificclimate/climate-explorer-frontend#12 right now. Should be done by end of day today/early tomorrow.

Issue #15, opened 2016-01-15
Some calls to data endpoint return all zeros


basilveerman:
Some data api calls are still returning incorrect information.

Some of the requests that @jameshiebert and I noted as returning zeros now return [expected](http://docker1.pcic.uvic.ca:9000/api/data?model=CanESM2&variable=rx5dayETCCDI&emission=rcp85&area=&time=1) [data](http://docker1.pcic.uvic.ca:9000/api/data?model=CSIRO-Mk3-6-0&variable=rx1dayETCCDI&emission=rcp26&area=POLYGON+%28%28-128.90625+60.41015624999999%2C+-128.90625+63.73046874999999%2C+-121.875+63.73046874999999%2C+-121.875+60.41015624999999%2C+-128.90625+60.41015624999999%29%29&time=0), but other still don't.

Using a constant polygon:

``` bash
POLYGON="POLYGON+((-127.9296875+45.76171875%2C+-127.9296875+54.6484375%2C+-118.65234375000001+54.6484375%2C+-118.65234375000001+45.76171875%2C+-127.9296875+45.76171875))"
```

Works with January:

``` bash
curl -s 'http://docker1.pcic.uvic.ca:9000/api/data?model=CSIRO-Mk3-6-0&variable=rx5dayETCCDI&emission=rcp85&area='$POLYGON'&time=0' | python -m json.tool
```

``` json
{
    "r1i1p1": {
        "data": {
            "2025-01-16T00:00:00Z": 94.53663330078125,
            "2055-01-16T00:00:00Z": 101.20841064453126,
            "2085-01-16T00:00:00Z": 111.72054443359374
        },
        "units": "mm"
    },
    "r2i1p1": {
        "data": {
            "2025-01-16T00:00:00Z": 98.94828491210937,
            "2055-01-16T00:00:00Z": 102.28649291992187,
            "2085-01-16T00:00:00Z": 114.10343017578126
        },
        "units": "mm"
    },
    "r3i1p1": {
        "data": {
            "2025-01-16T00:00:00Z": 93.01469116210937,
            "2055-01-16T00:00:00Z": 103.7731689453125,
            "2085-01-16T00:00:00Z": 112.4370849609375
        },
        "units": "mm"
    },
    "r4i1p1": {
        "data": {
            "2025-01-16T00:00:00Z": 95.85071411132813,
            "2055-01-16T00:00:00Z": 102.0443115234375,
[...]

PR #18, opened 2016-01-26
Adds "ensemble_name" and "timescale" as data request params

jameshiebert:
As mentioned in #15 , the `data` API call should filter out ensembles that are not a part of the application configuration. Additionally, it's possible for datasets to be presented in different time resolutions. For example, one might want to present downscaled climate model output as climatologies (e.g. 17 time steps or time_resolution='other') or as an annual timeseries (e.g. 150 or so timesteps). To do this, one must be able to specify the time_resolution when requesting data.

This PR adds both of those parameters to the `data` request and uses them to filter the available datasets from the model metadata database.

### DIFF
--- a/ce/api/data.py
+++ b/ce/api/data.py
@@ -3,10 +3,11 @@
 
 import numpy as np
 
-from modelmeta import Run, Time, Emission, Model, TimeSet, DataFile, DataFileVariable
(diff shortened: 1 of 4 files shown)


mdfischer:
With help from Basil we constructed some SQL queries to mimic what the code does, and generated  query strings to confirm the correct results being returned.  Everything seems to check out.

PR #20, opened 2016-03-17
Generate annual averages

mdfischer:
If you want to view some output files already produced by this version of generate_annual_averages, go to: /home/mfischer/code/climate-explorer-backend/data_prep/output_cheetah/

I will leave it to you to test write_time_dimension.


data_prep/generate_annual_averages.py
```diff
@@ -0,0 +1,191 @@
...
+    Generates file with annual averages for the given variable across
+    all time within an input file
+
+    Parameters:
+        in_fp: input file path
+        out_fp: output file path
+        variable (str): name of the variable which is being processed
+
+    '''
+    supported_vars = {
```
jameshiebert: The script is irrelevant to all of the ClimDEX variables since they are all defined as being annual.

data_prep/generate_annual_averages.py
```diff
@@ -0,0 +1,191 @@
...
+    if variable not in supported_vars:
+        raise Exception("Unsupported variable: can't yet process {}".format(variable))
+
+    if not os.path.exists(os.path.dirname(out_fp)):
+        os.makedirs(os.path.dirname(out_fp))
+
+    cdo = Cdo()
+    cdo.yearmean(input=in_fp, output=out_fp)
+
+def update_annual_avg_file_global_metadata(nc):
```
jameshiebert: One of my big pet peeves is naming a function or a task "update [something or other]". It really doesn't provide any description as to what the function is supposed to do or what the expected result is. One may as well name the function `randomly_change_global_metadata()` :) because all we know from the name is that something will be different afterwards.

data_prep/generate_annual_averages.py
```diff
@@ -0,0 +1,191 @@
...
+    # bounds_var might be a np.masked_array, and if any elements are masked, it's bad
+    if (hasattr(bounds_var, 'mask') and (np.ma.is_masked(bounds_var[:]))):
+        return False
+    # or, bounds_var might be full of zeros, also bad
+    elif (not np.any(bounds_var)):
+        return False
+    else:
+        return True
+
+def update_annual_avg_file_time_metadata(nc, start_year=None):
```
jameshiebert: I feel like the time var and the time_bnds var should be split out into two separate functions here...

data_prep/generate_annual_averages.py
```diff
@@ -0,0 +1,191 @@
...
+        .format(vars)))
+
+    if args.dry_run:
+        for f in test_files:
+            print(f)
+        sys.exit(0)
+
+    for fp in test_files:
+        log.info('Processing input file: {}'.format(fp))
+        file_ = Cmip5File(fp, freq='yr', mip_table='yr')
```
jameshiebert: I don't really like the use of the `Cmip5File` class here. Let's say we just want to test out the script on a single file in $TMPDIR. We can't actually do that, w/o creating a 9-level deep directory structure with a bunch of information that's already in the file attributes. E.g.

```python
(env34)hiebert@aether:~/code/climate-explorer-backend/data_prep$ python generate_annual_averages.py  -b /tmp/ -v tasmax -o /tmp/
INFO:__main__:Processing input file: /tmp/pr_year_BCCAQ+ANUSPLIN300+PCIC12_historical+rcp85_r1i1p1_1950-2100.nc
Traceback (most recent call last):
  File "generate_annual_averages.py", line 193, in <module>
    main(args)
  File "generate_annual_averages.py", line 160, in main
    file_ = Cmip5File(fp, freq='yr', mip_table='yr')
  File "/home/hiebert/code/climate-explorer-backend/data_prep/Cmip5File.py", line 29, in __init__
    self.institute, self.model, self.experiment, self.freq, self.realm, self.mip_table, self.run, self.version, self.variable = splitdirs[-9:]
ValueError: need more than 2 values to unpack
```
It makes the usage of the script too rigid. Is there really anything in that class that buys us anything?

mdfischer: I guess the intent was to conform to the CMIP5 standard for output file paths.  I guess a less rigid option would be to extract the metadata from the source NetCDF itself.

PR #25, opened 2016-04-12
Addresses code quality

jameshiebert:
These commits make a variety of surface-level code quality changes to increase our Landscape.io "health". Should be pretty straightforward.

### DIFF
--- a/ce/api/data.py
+++ b/ce/api/data.py
@@ -3,7 +3,7 @@
 
 import numpy as np
 
-from modelmeta import Run, Time, Emission, Model, TimeSet, DataFile
+from modelmeta import Run, Emission, Model, TimeSet, DataFile
 from modelmeta import DataFileVariable, EnsembleDataFileVariables, Ensemble
 from ce.api.util import get_array, get_units_from_run_object, get_files_from_run_variable, open_nc
 
--- a/ce/api/geo.py
+++ b/ce/api/geo.py
@@ -1,5 +1,4 @@
 import logging
-import collections
 import sys
 import math
 from collections import OrderedDict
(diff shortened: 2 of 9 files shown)


basilveerman:
These all look like good practices to follow.... nearly :P. I think all the `%` log message format recommendations are actually a misinterpretation of [PEP-282](https://www.python.org/dev/peps/pep-0282/). It does say to use % formatters in the log format definition like `%(name)s %(level)s - %(message)s`, but it does not specify how you should pass in a log message when *using* a logger instance.

jameshiebert:
Well, the [logging module docs](https://docs.python.org/3.4/howto/logging-cookbook.html#formatting-styles) discusses the fact that using a logging instance (i.e. calls to `debug()`, `info()`, etc.), can only take positional arguments for the log message, because kwargs are used for contextual information, and the logging module uses % formatting style internally. That doesn't really prevent using the style that we have been (debug('{} {}'.format()), but I think the reasoning behind Landscape's (and by extension pylint) marking is sound.

The idea is that you're supposed to push the interpolation to the logging framework rather than doing so in your code. This pushes the string formatting overhead to potentially be outside of your application, and avoids the possibility of raising exceptions during the formatting. Also, explicitly using .format() inside the logging call, the formatting will *always* be evaluated, even if your logging level precludes actually logging that message. Not so when passing parameters as arguments. These advantages are mentioned [here](https://www.simonmweber.com/2014/11/24/python-logging-traps.html).

So, long story short, from what I've read, that recommendation (pylint 1202) isn't out of line with best practices.

basilveerman:
Thanks for the clarification!

PR #28, opened 2016-04-13
Adds a more descriptive 400 error message

jameshiebert:
API requests that are missing required query parameters error out
with an HTTP 400 error. The message has been entirely unhelpful:
"Missing query parameters". This commit adds the names of the
missing query parameters to the error message so the user has some path
forward.

### DIFF
--- a/ce/tests/test_api.py
+++ b/ce/tests/test_api.py
@@ -1,5 +1,6 @@
 import math
 from time import time
+import re
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


jameshiebert:
@rod-glover I think that this is pretty straightforward, but would you mind taking a look at it for me?

ce/tests/test_api.py
```diff
@@ -23,6 +23,22 @@ def test_api_endpoints_are_callable(test_client, cleandb, endpoint, query_params
...
+def test_missing_query_params(test_client, cleandb):
+    url = '/api/data'
+    response = test_client.get(url)
+    assert response.status_code == 400
+    content = response.data.decode(response.charset)
+    assert "Missing query params:" in content
+    # Ensure that the response *tells* you what's wrong!
+    for param in ('model', 'emission', 'time', 'area', 'variable'):
+        assert param in content
+
```
rod-glover: Could DRY these 2 tests up into 1 by parametrizing over `url` and the list of expected param names in message.

rod-glover:
LGTM other than tightening up test code.

ce/tests/test_api.py
jameshiebert: Yeah, good call!

PR #29, opened 2017-03-02
Get all metadata from NetCDF header

rod-glover:
Modify existing generate_climos script to 

- take the list of files to be processed from the command line (only)
- get all metadata from the NetCDF header
- put PCIC standards-compliant metadata for climatology into the output files
- generate PCIC standards-compliant output file names
- optionally create one output file per dependent variable in the input file
- handle a wider range of input file types (GCM output, downscaled output, hydromodel output)
- use Python cdo functions a little more elegantly
- on dry-run, log a whole bunch o' info about the files to be processed
- put date/time in log messages

`generate_climos` also now has unit tests run against a selection of 'tiny' netCDF files representing the various file types this script may be run against. 

Following the original pattern of isolating `data_prep` from the rest of the CE backend, the `data_prep` package is independent of the `ce` package. This may or may not be a desirable state of affairs. We may want to refactoring so that `data_prep` is just a module of the CE backend.


data_prep/generate_climos.py
```diff
@@ -5,37 +5,22 @@
...
 from datetime import datetime
-from tempfile import NamedTemporaryFile
 
 from cdo import Cdo
-from netCDF4 import Dataset, num2date, date2num
+from netCDF4 import date2num
 from dateutil.relativedelta import relativedelta
 
-from Cmip5File import Cmip5File, ClimdexFile
+from util import s2d, ss2d, d2ss, d2s
```
jameshiebert: These are now in `nchelpers`, right? Should they be exposed/exported from `nchelpers` or is the fact that we need them here indicative that we haven't accomplished proper encapsulation in `nchelpers`? Or... *do* we even need them here?

data_prep/generate_climos.py
```diff
@@ -52,34 +37,20 @@ def iter_matching(dirpath, regexp):
...
-        t_end (datetime.datetime): end date of climo period
-        variable (str): name of the variable which is being processed
+    Parameters:
+        outdir (str): path to base directory in which to store output climo file
+        input_file (nchelpers.CFDataset): the input data file
+        t_start (datetime.datetime): start date of climo period to process
+        t_end (datetime.datetime): end date of climo period to process
 
-    Requested date range MUST exist in the input file
+    Requested date range MUST exist in the input file.
```
jameshiebert: What happens if it doesn't? Does this code assert that this assumption has been met?

data_prep/generate_climos.py
```diff
@@ -52,34 +37,20 @@ def iter_matching(dirpath, regexp):
...
 
-    '''
+    """
+    # TODO: Are these all really legitimate variables for processing, i.e., for forming temporal means over?
```
jameshiebert: Yes, good question. Might be worth passing that ask on to Trevor.

data_prep/generate_climos.py
```diff
@@ -90,45 +61,43 @@ def create_climo_file(fp_in, fp_out, t_start, t_end, variable):
         'tasmax', 'pr'
     }
 
-    if variable not in supported_vars:
-        raise Exception("Unsupported variable: cant't yet process {}".format(variable))
-
-    # Allow different ops by variable? # op = 'sum' if variable == 'pr' else 'mean'
```
jameshiebert: Are we sure that this is the case?

data_prep/generate_climos.py
```diff
@@ -90,45 +61,43 @@ def create_climo_file(fp_in, fp_out, t_start, t_end, variable):
...
+        # Premultiply input values by 86400
+        temporal_subset = cdo.mulc('86400', input=temporal_subset)
+
+    # Process selected data into climatological means
+    def climo_outputs(time_resolution):
+        '''Return a list of cdo operators that generate the desired climo outputs.
+        Result depends on the time resolution of input file data - different operators are applied depending.
+        If operators depend also on variable, then modify this function to depend on variable as well.
+        '''
+        ops_by_resolution = {
```
jameshiebert: This is a nice way to handle this. More clear than the old code.

data_prep/generate_climos.py
```diff
@@ -189,77 +159,117 @@ def update_climo_time_meta(fp, file_type=Cmip5File):
...
-    FileType = ClimdexFile if args.climdex else Cmip5File
 
-    for fp in test_files:
-        log.info(fp)
+def main(args):
+    if args.basedir:
+        variables = '|'.join(args.variables)
+        filepaths = iter_matching(
+            args.basedir,
+            re.compile('.*({}).*_(historical)?((?<=l)\+(?=r))?(rcp26|rcp45|rcp85)?_.*r\di\dp\d.*nc'.format(variables))
```
jameshiebert: Personally, I feel like we should remove this regex and the iter_matching, and just have the script accept a variable list of files as arguments and then iterate of the files. We can use cmd line tools like `find` easily and it makes the tool much more flexible.

data_prep/generate_climos.py
```diff
@@ -189,77 +159,117 @@ def update_climo_time_meta(fp, file_type=Cmip5File):
...
+            filepaths = [l for l in (l.strip() for l in f) if l[0] != '#']
+    else:
+        filepaths = []
 
-        nc = Dataset(fp)
-        available_climo_periods = determine_climo_periods(nc)
-        nc.close()
-        file_ = FileType(fp)
-        variable = file_.variable
+    filepaths = list(filepaths)  # Not very nice, but we reuse the list, at least for now
```
jameshiebert: This would go away if we followed my suggestion above.

data_prep/generate_climos.py
```diff
@@ -189,77 +159,117 @@ def update_climo_time_meta(fp, file_type=Cmip5File):
...
+    logger.info('Will process the following files:')
+    for filepath in filepaths:
+        logger.info(filepath)
 
-            # Create climatological period and update metadata
-            log.info('Generating climo period %s to %s', d2s(t_range[0]), d2s(t_range[1]))
-            out_fp = file_.generate_climo_fp(t_range, args.outdir)
-            log.info('Output file: %s', format(out_fp))
+    if args.dry_run:
+        logger.info('DRY RUN')
```
jameshiebert: So, does the dry run just print out attribute information?

data_prep/generate_climos.py
rod-glover: Sorry, don't understand the question here.

data_prep/generate_climos.py
rod-glover: I'm good with that. I do find the capacity additionally to read the set of input files from a file is useful, especially given the extremely long filepaths when testing from a dev machine.

data_prep/generate_climos.py
rod-glover: Outsourced to nchelpers -- only one was in use

data_prep/generate_climos.py
rod-glover: Yes. Originally, it did less than that, just printed the paths of the files to be processed.

data_prep/generate_climos.py
```diff
@@ -1,85 +1,40 @@
...
+        input_file (nchelpers.CFDataset): the input data file
+        t_start (datetime.datetime): start date of climo period to process
+        t_end (datetime.datetime): end date of climo period to process
 
-def iter_matching(dirpath, regexp):
-    # http://stackoverflow.com/questions/4639506/os-walk-with-regex
-    """Generator yielding all files under `dirpath` whose absolute path
-       matches the regular expression `regexp`.
-       Usage:
+    Requested date range MUST exist in the input file.
```
rod-glover: Not entirely sure what happens if it doesn't. (This comment was in original script.) What I think happens is that the output file contains only a subset of that date range, and therefore the temporal part of the output filename would be incorrect. First thing that happens in processing is cdo.seldate, which selects a temporal subset of the data in the input file.

data_prep/generate_climos.py
jameshiebert: The question is, "is our operation always mean"? You have removed the comment that suggests that perhaps we want to use "sum" for precipitation, so can I assume that you clarified with the scientists that we always want to use mean in this context?

data_prep/generate_climos.py
jameshiebert: Sure, but reading a set of input files from a file is as easy as:
```bash
cat file_list.txt | xargs python my_script.py
```
No additional code needed.

data_prep/generate_climos.py
jameshiebert: So... shouldn't this be from ``nchelpers.date_util import d2s, d2ss``?

data_prep/generate_climos.py
rod-glover: Perhaps my transfer of this into the docstring for `climo_outputs` (line 69) wasn't clear enough. But I did not intend to remove that information.

That said, it's worth checking with scientists. Trevor is coming by to talk about processing the climdex variables, I'll add this to the agenda.

data_prep/generate_climos.py
rod-glover: We're in some kind of github time warp here. If you look on the Files Changed tab, you will see that is exactly what I did.

data_prep/generate_climos.py
rod-glover: Argh, of course.

data_prep/generate_climos.py
rod-glover: Trevor says it is legit to process monthly climdex variables into climatological means. However, some of these variables will already be in the form of climatological means, so we need to be sure we are not trying to reprocess these.

rod-glover:
Travis CI fails because it is running the now-existent tests for `data_prep`. The test fail for two reasons:

- the `data_prep` package is not installed
- at this moment, `nchelpers` has not been published in a version that supports this script; the necessary changes are waiting in a PR

data_prep/dp/generate_climos.py
```diff
@@ -0,0 +1,357 @@
...
+        seasonal: 4 seasons: DJF, MAM, JJA, SON
+        annual: 1 value
+    """
+
+    # Year of all time values is middle year of period
+    year = (t_start + (t_end - t_start)/2).year + 1
+
+    times = []
+    climo_bounds = []
+
```
rod-glover: Do the computations of climatological times for month and season really have to be this complicated? Examples in the [CF Metadata Standard](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#climatological-statistics) consistently use the 15th day of each month for multi-year monthly averages and the 16th day of the mid-season months (Jan, Apr, July, Oct) for multi-year seasonal averages. Any reason not to use these?

data_prep/dp/generate_climos.py
```diff
@@ -0,0 +1,357 @@
...
+        outdir (str): path to base directory in which to store output climo file(s)
+        input_file (nchelpers.CFDataset): the input data file
+        split_vars: (bool) if True, produce one file per dependent variable in input file;
+            otherwise produce a single output file containing all variables
+        t_start (datetime.datetime): start date of climo period to process
+        t_end (datetime.datetime): end date of climo period to process
+
+    We use CDO to perform the key computations for the climatology outputs.
+
+    One output file is generated per dependent variable in the input file. This simplifies the processing somewhat,
```
rod-glover: Variable splitting now optional.

data_prep/tests/test_create_climo_files.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+@mark.parametrize('input_and_climo_files, t_start, t_end', [
+    (('gcm', False, t_start(1965), t_end(1970)), t_start(1965), t_end(1970)),
+    (('downscaled_tasmax', False, t_start(1961), t_end(1990)), t_start(1961), t_end(1990)),
+    # No need to repleat with downscaled_pr
+    (('hydromodel_gcm', False, t_start(1984), t_end(1995)), t_start(1984), t_end(1995)),
+    (('hydromodel_gcm', True, t_start(1984), t_end(1995)), t_start(1984), t_end(1995)),
+], indirect=['input_and_climo_files'])
+def test_time_and_climo_bounds_vars(input_and_climo_files, t_start, t_end):
```
rod-glover: Good. And to repeat a comment I put in the code under test:

Do the computations of climatological times for month and season really have to be this complicated? Examples in the [CF Metadata Standard](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#climatological-statistics) consistently use the 15th day of each month for multi-year monthly averages and the 16th day of the mid-season months (Jan, Apr, July, Oct) for multi-year seasonal averages. Any reason not to use these?

rod-glover: I'd really like to use a simpler computation (and tests) if there is no reason to be so complicated.

data_prep/tests/test_create_climo_files.py
```diff
@@ -0,0 +1,205 @@
...
+            def d2n(date):
+                return date2num(date, cf.time_var.units, cf.time_var.calendar)
+
+            # Test monthly mean timesteps and climo bounds
+            if expected_num_time_values == 17:
+                for month in range(1, 13):
+                    t = next(time_steps)
+                    assert t.year == climo_year
+                    assert t.month == month
+                    assert t.day in [15, 16]
```
rod-glover: Yes. But see comment above re. complication.

data_prep/tests/test_create_climo_files.py
```diff
@@ -0,0 +1,205 @@
+"""Unit tests for the core function create_climo_files.
+
+These tests are all parametrized over several test input files, which requires a little trickiness with fixtures.
+pytest doesn't directly support parametrizing over fixtures (which here deliver the test input files
+and the output of create_climo_files). To get around that, we use indirect fixtures, which are passed a parameter
```
rod-glover: Bleh. Thought I had this explanation there too. I will transplant it.

setup.py
```diff
@@ -30,14 +30,20 @@ def find():
...
+        'nchelpers',
         'GDAL',
-        'rasterio'
+        'rasterio',
+        'cdo',
+        'pytest',
     ],
     scripts=['scripts/devserver.py'],
-    package_dir={'ce': 'ce'},
+    package_dir={  # necessary?
```
jameshiebert: Pretty sure that it's necessary.

requirements.txt
```diff
@@ -5,7 +5,11 @@ modelmeta
 shapely
 numpy
 netcdf4
-GDAL
+# nchelpers
+GDAL<2.0 # Allow pip or system install for travis
```
jameshiebert: Why < 2.0?

requirements.txt
```diff
@@ -5,7 +5,11 @@ modelmeta
...
 rasterio
+cdo
 # For testing
+pytest
 python-dateutil
+pip
```
jameshiebert: pip is kind of assumed to exist. Should not be included in requirements.txt. I.e. how would you "pip install -r requirements.txt" if you didn't have pip installed?

rod-glover:
@jameshiebert 

Test data files are all < 1Mb, except for `tiny_hydromodel_gcm.nc`, which is more than twice as large as I would expect it to be given its very modest dimensions (4383 time steps, 2 x 2 grid, 3 levels). There are 6 vars, but that doesn't account for the large size, which I estimate should be about 1.5 Mb.

All data prep tests pass, if the most up-to-date version of `nchelpers` is installed. That version is in PR now, not yet on our PyPI.

Let me know if you are happy to merge this PR as is.

jameshiebert:
Yeah, I don't get how it's 4MB either... should only be 1.6MB if it's using double precision float. Weird. We'll live, so let's move on.

PR #32, opened 2017-05-10
Add instructions for installing and using generate_climos

rod-glover:
`generate_climos` will typically be used stand-alone on a compute node. It's nice to have installation instructions; installing `climate-explorer-backend` is not a useful option here.

This update also contains an example bash script (actually used) for the standard process of: copy input to temp, run script, copy output to permanent storage.

### DIFF
--- a/dp/process-climo-means.sh
+++ b/dp/process-climo-means.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# process a model output file into climatological means
+# - copy input file to local temp
+# - process it with generate_climos.py
+# - copy results from local temp to permanent location
+# - delete copy of input file
+#
+# arguments:
+# $1: unique id for logfile so that this script can be run in several independent processes. Easy: Pass in $$.
+# $2: path to file to be processed
+#
+# typical usage:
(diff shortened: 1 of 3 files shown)


dp/process-climo-means.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# process a model output file into climatological means
+# - copy input file to local temp
+# - process it with generate_climos.py
+# - copy results from local temp to permanent location
+# - delete copy of input file
+#
+# arguments:
+# $1: unique id for logfile so that this script can be run in several independent processes. Easy: Pass in $$.
+# $2: path to file to be processed
```
jameshiebert: It might be useful if the remaining arguments were a *list* of filenames to be processed. That way you can do the standard ``find -name *.nc | xargs process-climo-means.sh logfile.txt``.

To do this in the script, you would consume $1 as the logfile and then use the bash ``shift`` operation, and then the remaining args are available in $@ (I think).

dp/requirements.txt
```diff
@@ -0,0 +1,4 @@
+# requirements for data prep (dp) component alone
```
jameshiebert: Good call.

PR #33, opened 2017-05-10
Add update-metadata utility

rod-glover:
This utility script reads changes to be made from a specified YAML file and applies them to the specified NetCDF file. Updates possible are:

- set global attribute
- delete global attribute
- rename global attribute

Unfortunately, the initial cut at this utility was merged into master some while ago. This PR just fixes a bug and improves the code a little.


jameshiebert:
Yep, that should work!

Issue #34, opened 2017-05-10
generate_climos fails on files with 360_day calendar


rod-glover:
Example: Processing `/storage/data/climate/CMIP5/output/NIMR-KMA/HadGEM2-AO/historical/day/atmos/day/r1i1p1/v20130422/tasmin/tasmin_day_HadGEM2-AO_historical_r1i1p1_18600101-20051230.nc`:

```2017-05-09 22:02:57 INFO: Processing: /local_temp/rglover/tasmin_day_HadGEM2-AO_historical_r1i1p1_18600101-20051230.nc
2017-05-09 22:02:58 INFO: Generating climo period 1961-01-01 to 1990-12-30
2017-05-09 22:02:58 INFO: Selecting temporal subset
2017-05-09 22:03:10 INFO: Forming climatological means
Traceback (most recent call last):
  File "generate_climos.py", line 372, in <module>
    main(args)
  File "generate_climos.py", line 356, in main
    convert_longitudes=args.convert_longitudes, split_vars=args.split_vars)
  File "generate_climos.py", line 179, in create_climo_files
    update_climo_time_meta(climo_means)
  File "generate_climos.py", line 314, in update_climo_time_meta
    climo_bnds_var[:] = date2num(climo_bounds, cf.time_var.units, cf.time_var.calendar)
  File "netCDF4/_netCDF4.pyx", line 5265, in netCDF4._netCDF4.date2num (netCDF4/_netCDF4.c:65355)
  File "netcdftime/_netcdftime.pyx", line 795, in netcdftime._netcdftime.utime.date2num (netcdftime/_netcdftime.c:14358)
ValueError: there are only 30 days in every month with the 360_day calendar
```

This occurs for all model output files under `/storage/data/climate/CMIP5/output/NIMR-KMA/` and `/storage/data/climate/CMIP5/output/MOHC/`.

rod-glover:
Problem is in climo_bounds calculation of annual mean end-date

PR #35, opened 2017-05-10
Fix climo bounds calculation for 360 day calendar

rod-glover:
This fixes https://github.com/pacificclimate/climate-explorer-backend/issues/34

New annual average climo bound end date is computed as first day of next year instead of end date + 1 day. This is correct so long as the assumption that climatological periods are full years remains correct. Pretty much guaranteed given the definition of monthly climo means.


jameshiebert:
Sometimes climatologies for hydrologic values are computed based on the "hydrologic calendar", which varies depending on who has defined it. I think that Water Survey Canada uses Oct 1 - Sept 30? So, you may need to be careful with this assumption.

rod-glover:
Hmm. This poses some additional questions/issues:

- Start and end dates for forming the means are currently fixed on Jan 1 boundaries. This apparently needs to be modified to accommodate hydrology.
- What difference, if any, does the hydrologic calendar make to the notion of seasons? (And I just realized we have a slight issue with the winter season on Jan 1-bounded years: The first winter technically begins one month before the start of the specified period, so its mean is necessarily slightly impaired. Fingers crossed that CDO does a sensible thing here and does not divide a 2-month sum by 3 or something.)
- Happily, I think that the "annual" mean -- mean over years of annual mean -- doesn't differ, or differs by vanishingly little. Script calculates it by taking the mean over all values in the climatological period, so the notion of "year" doesn't actually enter into it.

rod-glover:
Met with @mschnorb re. the effect of the "hydrological year" starting on a non-Jan boundary. Conclusion is that the difference this would make to seasonal and annual means is negligibly small and need not be accounted for in our code. Climatological means can safely be taken over 30 year periods that begin and end at Jan 1.

I will add a comment to this effect in the code that calculates begin and end dates.

rod-glover:
Which is in `nchelpers`. I'll add a comment in an appropriate place in this code as well.

PR #37, opened 2017-05-11
Improve update_metadata

rod-glover:
Add PyYAML dependency
Enable updates to per-variable metadata as well as global metadata
Add documentation in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,14 +202,18 @@ You only need to know a couple of things about YAML and how we employ it to use 
   * The second (indented) level specifies the attribute and the change to be made to it.
     See below for details.
 
-##### Delete attribute:
+##### Delete attribute
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -179,3 +180,92 @@ index.netcdf.files.sqlite(f.list, "/tmp/mddb.sqlite3")
...
+
+Rename the attribute named `oldname` to `newname`. Value is unchanged.
+
+```yaml
+global-or-variable-name:
+    newname: <-oldname
+```
+
+Note: The special sequence `<-` after the colon indicates renaming. 
+This means that you can't set an attribute with a value that begins with `<-`. Sorry.
```
jameshiebert: Heh. I think we can live with this. :)

Issue #36, opened 2017-05-11
generate_climo scales precip units incorrectly for packed files 


rod-glover:
`generate_climo` can (under control of a flag) convert precipitation units from per-second to per-day. Currently this conversion is done by multiplying by a scale factor of 86400 (s/day). This is correct for absolute (unpacked) files, but fails for packed files where the offset value is non-zero.

Solution is to recognize packed files and apply a different computation to them taking into account the scale and offset values.

Question: Is there a standard for how packed files are defined that we can adhere to, or is it ad hoc?

jameshiebert:
> Question: Is there a standard for how packed files are defined that we can adhere to, or is it ad hoc?

I wrote a [blog post](http://james.hiebert.name/blog/work/2015/04/18/NetCDF-Scale-Factors/) a couple of years ago about the packing. It contains more information than you need, but it addresses some of the historic reasons for the ad-hoc nature of it, and also has a link to Unidata's best practices for packing.

But essentially, you should assume that it's ad-hoc and your code should take care.

rod-glover:
Happily, [Unidata's best practices for packing](http://www.unidata.ucar.edu/software/netcdf/docs/BestPractices.html) do indicate standard names for the packing param attributes: `scale_factor` and `add_offset`. Those are in use in files Stephen Sobie is preparing. 

I'll treat this (plus integer data type for the variable) as the indicator for packed values; any other case will get treated as not packed, correctly or no. Possibly could raise an error if the variable data type is an integer but the expected attributes are not present, although this could conceivably be incorrect in some places.

rod-glover:
Another question here is whether it is worth putting this effort in at the moment. It is only needed when precip needs to be scaled to different units, and Stephen says that his will always be in the desired units. I don't like leaving holes like that, but if we never use it, it might not be worth the effort to fix. As a placeholder, I could raise an exception when this condition is detected.

rod-glover:
However, the procedure for units conversion (multiplication by a constant factor) is actually **simpler** in the case of packed values: To multiply the unpacked values by a constant value `K`, multiply both the scale factor and offset by `K`. The packed values do not change.

jameshiebert:
> Stephen says that his will always be in the desired units

Well... many, many of the files that this software will service will not have been created personally by Stephen. We definitely need to cover this case.

> I'll treat this (plus integer data type for the variable) as the indicator for packed values

[...]

PR #38, opened 2017-05-12
Handle packed files

rod-glover:
This addresses #36


dp/generate_climos.py
```diff
@@ -134,19 +134,34 @@ def climo_outputs(time_resolution):
     # Apply per-variable processing
     # - Scale pr variable if it is not in desired units
     #   - TODO: Use UDUNITS here for more robust processing of units? It has a fairly heavy API. But this is uuuugly.
```
jameshiebert: You may want to check out the [``pint`` library](https://pint.readthedocs.io/en/0.7.2/) for units handling.

dp/generate_climos.py
```diff
@@ -134,19 +134,34 @@ def climo_outputs(time_resolution):
...
+    pr_attributes = {}  # will contain updates, if any, to pr variable attributes
     if 'pr' in input_file.dependent_varnames:
```
jameshiebert: I know that you didn't create this problem, but I don't like the big, long branch here for `pr` specific code. Could we stash that away in a subroutine or something to help this read better for the common cases?

dp/generate_climos.py
```diff
@@ -134,19 +134,34 @@ def climo_outputs(time_resolution):
...
-            # Extract variable
-            pr_only = cdo.select('name=pr', input=climo_means)
-            # Multiply values by 86400 to convert from mm/s to mm/day
-            pr_only = cdo.mulc('86400', input=pr_only)
-            # Replace pr in all-variables file
-            climo_means = cdo.replace(input=[climo_means, pr_only])
             # Update units attribute
-            # TODO: Verify that "d-1" is desired way to express "per day" (alternaive is "day-1")
-            pr_units_attr = re.sub('(/s|/ s|s-1|s\^-1)', ' d-1', units)
+            pr_attributes['units'] = re.sub('(/s|/ s|s-1|s\^-1)', ' d-1', units)
```
jameshiebert: I've seen s-1 for "per second", but I've never seen d-1 for "per day". Is that in the SI units?

dp/generate_climos.py
```diff
@@ -169,11 +184,9 @@ def climo_outputs(time_resolution):
         cf.climo_start_time = t_start.isoformat()[:19] + 'Z'
         cf.climo_end_time = t_end.isoformat()[:19] + 'Z'
         if hasattr(input_file, 'tracking_id'):
-            cf.climo_tracking_id=input_file.tracking_id
-        try:
-            cf.variables['pr'].units = pr_units_attr
-        except NameError:
```
jameshiebert: Silent errors are definitely bad, so I approve of getting rid of this. What actually happens here if the variable is not attributed with units? I guess we're assuming that everything going through this script passes nclint and therefore *is* attributed with units, eh? I'm fine with that, just making sure.

dp/generate_climos.py
rod-glover: What a delightful and elegant library! `pint` can't parse all valid `udunits` strings, but a simple transformation of the strings can fix that. 

Also the Python binding of `udunits2`, `cfunits`, only works in Python ver >=2.6 and <3.0, so we **must** use something different. I'm liking `pint`.

dp/generate_climos.py
rod-glover: 'd' (and for that matter 'day') is not an SI unit. But 'day' commonly used in extended units sets, such as those accepted by `udunits2` and `pint`. 'd' is recognized as a synonym for 'day' by `udunits2` (but not `pint`). 'd' is commonly used by scientists at PCIC.

dp/generate_climos.py
rod-glover: There is another Python binding of udunits2 to Python,[ `cf_units`](https://github.com/SciTools/cf_units). It appears to be Python 3 compatible. It's written by the UK Met office, and appears to be actively maintained. Upside: Exactly what is called for in this situation. Downside: Have to install `udunits2` separately.

dp/generate_climos.py
rod-glover: An answer in two parts:

1. The original try-catch block was just a shitty way decide if there were any updates to be done. Much better now.
2. If the variable is not actually attributed with units, the code will raise an AttributeError exception from netcdf. But we don't yet have an nclint check that all variables are attributed with units. (And we could probably add some other attributes to check as well.)

jameshiebert:
To me, this all looks good to go (except for the one line conflict).

Issue #45, opened 2017-06-17
generate_climos.py should create one file per climo averaging period


rod-glover:
Add an option in `generate_climos.py `to output one file per climatological averaging period (month, season, year).

Currently they are all in one file, with a single variable with a non-monotonic time dimension (12 months, 4 seasons, 1 year). This is a problem: ncWMS2 won't accept files with non-monotonic dimensions. Separating into different files solves this, and is formally more correct (the different averaging periods technically are distinct random variables).

Issue #42, opened 2017-06-02
TravisCI doesn't run data prep tests


rod-glover:
Explanation in `.travis.yml`: 

We only run the ce tests because installing cdo is just too hard: NetCDF support isn't built into the
debian cdo pacakges. Once that is resolved (i.e., install the NetCDF support before installing cdo),
we [can] run the data prep tests too.

rod-glover:
This problem has migrated to https://github.com/pacificclimate/climate-explorer-data-prep

Issue #39, opened 2017-05-12
Timing tests fail intermittently on Travis CI


rod-glover:
Several tests check time of execution: [ce/tests/test_api.py#L194](https://github.com/pacificclimate/climate-explorer-backend/blob/master/ce/tests/test_api.py#L194), [ce/tests/test_util.py#L39](https://github.com/pacificclimate/climate-explorer-backend/blob/master/ce/tests/test_util.py#L39), [ce/tests/test_geo.py#L40](https://github.com/pacificclimate/climate-explorer-backend/blob/master/ce/tests/test_geo.py#L40).

These tests intermittently fail on Travis CI, with execution times from a few percent to several hundred percent larger than asserted. Should not have restarted a recent failing build so I could link to it. Trust me.

rod-glover:
`push` test today failed by 2 orders of magnitude: https://travis-ci.org/pacificclimate/climate-explorer-backend/builds/245150786?utm_source=github_status&utm_medium=notification

The same test, for `pr`, succeeded: https://travis-ci.org/pacificclimate/climate-explorer-backend/builds/245151176?utm_source=github_status&utm_medium=notification

PR #47, opened 2017-06-20
generate_climos: Option to generate one file per averaging interval

rod-glover:
Resolves https://github.com/pacificclimate/climate-explorer-backend/issues/45


rod-glover:
Update README

dp/generate_climos.py
```diff
@@ -306,36 +296,72 @@ def convert_pr_var_units(input_file, climo_means):
...
+                12: {'monthly'},
+                4: {'seasonal'},
+                1: {'annual'},
+                5: {'seasonal', 'annual'},
+                17: {'monthly', 'seasonal', 'annual'},
+            }[cf.time_var.size]
         except KeyError:
-            raise ValueError("Climatology file must have a frequency value in {}; found '{}'"
-                             .format(frequency_to_time_types.keys(), cf.frequency))
+            raise ValueError()
```
rod-glover: Add an informative message to the exception

dp/generate_climos.py
```diff
@@ -306,36 +296,72 @@ def convert_pr_var_units(input_file, climo_means):
...
-def update_climo_time_meta(filepath):
+def split_on_variables(climo_means_file, var_names):
+    if len(var_names) > 1:
+        return [cdo.select('name={}'.format(var_name), input=climo_means_file)
+                for var_name in var_names]
+    else:
+        return [climo_means_file]
+
+
+def update_climo_metadata(input_file, t_start, t_end, climo_filepath):
```
rod-glover: Rename to `update_metadata_and_time_var`

dp/generate_climos.py
```diff
@@ -118,72 +130,50 @@ def climo_outputs(time_resolution):
...
+        for climo_means_file in climo_means_files:
+            convert_longitude_range(climo_means_file)
+
+    # Convert units on any pr variable in each file
+    climo_means_files = [convert_pr_var_units(input_file, climo_mean_file) for climo_mean_file in climo_means_files]
+
+    # Update metadata in climo files
+    logger.debug('Updating climo metadata')
+    for climo_means_file in climo_means_files:
+        update_climo_metadata(input_file, t_start, t_end, climo_means_file)
```
rod-glover: Better if  `update_climo_metadata` returned the file name, and we use a list comprehension to invoke it, so that if its implementation changes (e.g., to use CDO) this code will be robust.

dp/generate_climos.py
```diff
@@ -118,72 +130,50 @@ def climo_outputs(time_resolution):
...
-
-    # Update time metadata in climo file
-    update_climo_time_meta(climo_means)
-
-    # Create final output file(s): split climo means file by dependent variables if required
+    # Optionally convert longitudes in each file
+    if convert_longitudes:
+        logger.info('Converting longitudes')
+        for climo_means_file in climo_means_files:
+            convert_longitude_range(climo_means_file)
```
rod-glover: Better if `convert_longitude_range` returned the file name, and we use a list comprehension to invoke it, so that if its implementation changes (e.g., to use CDO) this code will be robust.

dp/generate_climos.py
```diff
@@ -398,15 +416,20 @@ def main(args):
...
-    parser.add_argument('-s', '--split-vars', dest='split_vars', action='store_true')
+    parser.add_argument('-g', '--convert-longitudes', dest='convert_longitudes', action='store_true',
+                        help='Transform longitude range from [0, 360) to [-180, 180)')
+    parser.add_argument('-v', '--split-vars', dest='split_vars', action='store_true',
+                        help='Generate a separate file for each dependent variable in the file')
+    parser.add_argument('-i', '--split-intervals', dest='split_intervals', action='store_true',
+                        help='Generate a separate file for each climatological period')
     parser.add_argument('-o', '--outdir', required=True, help='Output folder')
-    parser.set_defaults(dry_run=False, convert_longitudes=False, split_vars=False)
+    parser.set_defaults(dry_run=False, convert_longitudes=False, split_vars=False, split_intervals=True)
```
rod-glover: Oops, that should be False!

dp/tests/conftest.py
```diff
@@ -3,104 +3,12 @@
...
-    """Returns one of the input files, based on the request param.
-
-    request.param: (str) selects the input file to be processed by create_climo_files
-    returns: (nchelpers.CFDataset) input file to be processed by create_climo_files
-    This fixture should be invoked with indirection.
-    """
-    return get_input_file(request.param)
+@fixture
+def tiny_dataset(request):
+    return CFDataset(resource_filename('dp', 'tests/data/tiny_{}.nc').format(request.param))
```
rod-glover: Holy shit, what was I thinking before?

rod-glover:
@jameshiebert , this is a blocker to me generating more climo means. There are a bunch of downscaled files ready to go now (and they will each take a while).

rod-glover:
Failing test(s) is transient problem documented in https://github.com/pacificclimate/climate-explorer-backend/issues/39

rod-glover:
Also note that Travis does not run tests for data prep, as per https://github.com/pacificclimate/climate-explorer-backend/issues/42. But all tests pass in my dev environment.

jameshiebert:
Does any of this functionality require us to roll a new version of nchelpers?

dp/generate_climos.py
```diff
@@ -398,15 +419,20 @@ def main(args):
...
-#    parser.add_argument('-c', '--climo', nargs= '+',  help='Climatological periods to generate. IN PROGRESS. Defaults to all available in the input file. Ex: -c 6190 7100 8100 2020 2050 2080')
+    # parser.add_argument('-c', '--climo', nargs= '+',  help='Climatological periods to generate.
+    # IN PROGRESS. Defaults to all available in the input file. Ex: -c 6190 7100 8100 2020 2050 2080')
     log_level_choices = 'NOTSET DEBUG INFO WARNING ERROR CRITICAL'.split()
     parser.add_argument('-l', '--loglevel', help='Logging level',
                         choices=log_level_choices, default='INFO')
     parser.add_argument('-n', '--dry-run', dest='dry_run', action='store_true')
-    parser.add_argument('-g', '--convert-longitudes', dest='convert_longitudes', action='store_true')
-    parser.add_argument('-s', '--split-vars', dest='split_vars', action='store_true')
+    parser.add_argument('-g', '--convert-longitudes', dest='convert_longitudes', action='store_true',
```
jameshiebert: One thing that I notice about all of these arguments (convert-longitudes, split-vars and split-intervals), is that they are False by default. However each of them are essentially represent "do something that is the best practice. So by default we're doing the *wrong* thing, and making the user explicitly ask the script to do the *right* thing.

Changing the behaviour  would probably mean changing the default to ``True``, removing the ``action`` parameter, changing the type to bool and making the user say something like ``--split-vars=False``.

dp/generate_climos.py
```diff
@@ -306,36 +296,74 @@ def convert_pr_var_units(input_file, climo_means):
...
-    :param filepath: (str) filepath to a climatological means output file which needs to have its
-        time variable
+    :param input_file: (CFDataset) input file from which the climatological output file was produced
+    :param t_start: (datetime.datetime) start date of climatological output file
+    :param t_end: (datetime.datetime) end date of climatological output file
+    :param climo_filepath: (str) filepath to a climatological means output file which needs to have its
+        metadata update
 
-    IMPORTANT: THIS MAKES CHANGES TO FILES IN PLACE
+    WARNING: THIS CHANGES FILE `climo_filepath` IN PLACE
```
rod-glover: That is directed at the developer. It's in distinction to the CDO operators, which always generate a new file containing the modification. In-place doesn't matter much unless you apply it directly to an input file. In this code, it is always applied to an intermediate results file.

dp/generate_climos.py
rod-glover: Agree entirely. Will do.

dp/tests/conftest.py
rod-glover: Well, it upped my chops with pytest fixtures, for sure.

rod-glover:
This PR does not rely on a new version of nchelpers.

rod-glover:
Addressed all comments.

jameshiebert:
W00t!

PR #49, opened 2017-06-24
Scripts for submitting generate_climo jobs to PBS

rod-glover:
NEW & SHINY

Python scripts for managing and using a database containing a queue of files to be processed with `generate_climos`. Defined in the spirit of the PBS commands `qsub`, `qstat`, etc. 

Note there are now two queues in play: the `generate_climos` queue, and the PBS job queue. This tool mediates between them.

Of note:
- `gclist.py`: Lists contents of `generate_climos` queue.
- `gcadd.py`: Adds a file to be processed (with processing options) to the `generate_climos` queue.
- `gcsub.py`: Submits the next N unsubmitted files in the `generate_climos` queue to PBS for processing. Generates a customized PBS script for each submission.
- `gcalter-params.py`: Alters generate_climos params and PBS params of queue entries, so that fixes/changes can be made after queueing. (Does not allow for changes after submission to PBS queue.)
- `gcupd-email.py`: Updates the status of a submitted  `generate_climos` queue entry using the contents of a PBS status email.
- `gcupd-qstat.py`: Updates the statuses of all submitted `generate_climos` queue entries by invoking `qstat` for each one and parsing the result.
- Uses [Alembic](http://alembic.zzzcomputing.com/en/latest/index.html), a widely used database migration manager for SQLAlchemy, to manage database schema. This is a very simple database, and thus the perfect place to trial it. Early wins: 
    - First defined the database with SQLAlchemy, then used Alembic to auto-generated the database creation code and create the database. (Alembic can auto-generate any further migrations as well.)
    - Easy to configure and manage multiple test and prod databases. See `alembic.ini` and `env.py`.

-----------------------

OLD & TARNISHED (removed)

This comes in two pieces:

- `qsub-climo.sh`: Submit a job to the PBS queue that runs generate_climos on a single NetCDF file. 
  - We want a PBS job to process just one file to improve scheduling and $TMPDIR usage. 
- `qsub-climos-from-file.sh`: Submit generate_climos jobs to PBS queue ...
  - Read files to process from a 'todofile'. 
  - Write submitted files to a 'donefile' (with PBS job id identifier). Do not resubmt files found in donefile. 
  - This allows us to submit jobs incrementally with automatic bookkeeping, and to add new files to the todofile and process them without any extra effort.  
  - Submits using `qsub-climo.sh`


jameshiebert:
So I know that the file-based bookkeeping is relatively simple. However, in the past with jobs like this (essentially a task queue), we've used a simple sqlite database that has a `status` column ("NEW", "INPROGRESS", "DONE", "ERROR") or something like that. Then there is *one* definitive place to look for information (rather than having to compare multiple files and do set differentials on them). See [`esgf_download`](https://github.com/pacificclimate/esgf_download) for an example.

Is that something you'd be willing to consider for this?

rod-glover:
Certainly. I'll have a look at the example and let you know if I have any questions. But assume I'll go ahead on this.

rod-glover:
Follow-up question: Where do you want to house the database? That is, would you want it to be part of the climate-explorer-backend repo? Or housed somewhere else?

dp/jobqueueing/alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
rod-glover: Probably should move some or all of the explanation of multi-database configuration out of env.py into here.

rod-glover:
@jameshiebert , these scripts have been tested up to but not including submission locally (and work). I've also tested submission in production, with a few downscaled datasets. Jobs were submitted fine using `gcsub`, but it remains to be seen whether I made any errors in the `qsub` script it submits.

rod-glover:
If you want to examine the production queue: 
- install the backend on an interactive compute host
- `cd ./dp/jobqueueing` 
- `python gclist.py -d /storage/data/projects/comp_support/climate_exporer_data_prep/climatological_means/jobqueue-prod.sqlite` (I made a symlink to this file to reduce CLI effort.)

rod-glover:
Add some description of this tooling to README

dp/jobqueueing/gcalter-params.py
```diff
@@ -0,0 +1,70 @@
...
+logger = default_logger()
+
+
+def update_generate_climos_queue_entries_with_params(session, args):
+    entries = session.query(GenerateClimosQueueEntry)\
+        .filter(GenerateClimosQueueEntry.status == 'NEW')\
+        .all()
+
+    def args_to_entry(attr):
+        """Oh for *#@%* sake. Oughta migrate the database but SQLite doesn't rename columns easily.
```
jameshiebert: Oh boy...

dp/jobqueueing/gcreset.py
```diff
@@ -0,0 +1,96 @@
+"""
+Script to update the status of a queue entry.
```
jameshiebert: Can you be more specific (either here or in the `ArgumentParser` help message) about what it's updating the status to be? Or what information it's using to "reset" the status of a queue entry? In `main()` you mention using an e-mail message as input, but what's the intended workflow. Save e-mail to disk and then point this script to the output file? If so, it's not overwhelmingly clear without reading the code in-depth.

dp/jobqueueing/gcsub.py
```diff
@@ -0,0 +1,164 @@
...
+#PBS -m abe
+#PBS -N generate_climos:{input_filename}
+#PBS -d /storage/home/rglover/code/climate-explorer-backend
+
+pbs_job_num=$(expr match "$PBS_JOBID" '\([0-9]*\)')
+
+# Set up the execution environment
+module load netcdf-bin
+module load cdo-bin
+source py3.4/bin/activate
```
jameshiebert: This is fragile... unstated dependency of having a virtual environment installed in a specific path.

dp/jobqueueing/gcsub.py
```diff
@@ -0,0 +1,164 @@
...
+def make_qsub_test_script(queue_entry):
+    return '''
+#PBS -l nodes=1:ppn={qe.ppn}
+#PBS -l vmem={vmem}mb
+#PBS -l walltime={qe.walltime}
+#PBS -o {qe.output_directory}/logs
+#PBS -e {qe.output_directory}/logs
+#PBS -m abe
+#PBS -N generate_climos:{input_filename}
+#PBS -d /storage/home/rglover/code/climate-explorer-backend
```
jameshiebert: Can we create a secure TMPDIR for a working directory? Rather than using your $HOME dir (which won't work for anyone else).

dp/jobqueueing/gcsub.py
```diff
@@ -0,0 +1,164 @@
...
+def make_qsub_script(queue_entry):
+    return '''
+#PBS -l nodes=1:ppn={qe.ppn}
+#PBS -l vmem={vmem}mb
+#PBS -l walltime={qe.walltime}
+#PBS -o {qe.output_directory}/logs
+#PBS -e {qe.output_directory}/logs
+#PBS -m abe
+#PBS -N generate_climos:{input_filename}
+#PBS -d /storage/home/rglover/code/climate-explorer-backend
```
jameshiebert: Likewise.

dp/jobqueueing/gcsub.py
```diff
@@ -0,0 +1,164 @@
...
+echo infile = $infile
+
+# Output directory is automatically created by generate_climos
+baseoutdir=$TMPDIR/climo/output
+echo baseoutdir = $baseoutdir
+outdir=$baseoutdir/$pbs_job_num
+echo outdir = $outdir
+
+# Generate climo means
+python dp/generate_climos.py -g {qe.convert_longitude} -v {qe.split_vars} -i {qe.split_intervals} -o $outdir $infile
```
jameshiebert: This is nice how you have preserved the configurability from the main script up to the high-level operation of queueing a set of file.

dp/jobqueueing/gcsub.py
```diff
@@ -0,0 +1,164 @@
...
+# Since output files are small, we're not removing them here.
+rsync -r $baseoutdir {qe.output_directory}
+ls $baseoutdir
+rm $infile
+            '''.format(qe=queue_entry,
+                       input_filename=os.path.basename(queue_entry.input_filepath),
+                       vmem=queue_entry.ppn * 12000)
+
+
+def submit_generate_climos_pbs_jobs(session, args):
```
jameshiebert: You might consider splitting `args` out into its child components that are actually used in the function (`number` and `test_job`, I think).

Absent doing so, there's nearly nothing in this file to indicate what the real input to this function is (without tracing the function and reading each place that `args` is used). Since the options to the `ArgumentParser` are hidden away in the helpers module.

dp/jobqueueing/gcupd-qstat.py
```diff
@@ -0,0 +1,82 @@
...
+
+from dp.script_helpers import default_logger
+from dp.jobqueueing.argparse_helpers import add_global_arguments
+from dp.jobqueueing.jobqueueing_db import GenerateClimosQueueEntry
+
+
+logger = default_logger()
+
+
+def qstat_to_yaml(string):
```
jameshiebert: That's creative :)

dp/jobqueueing/jobqueueing_db.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    __tablename__ = 'generate_climos_queue'
+    id = Column('generate_climos_queue_id', Integer, primary_key=True, nullable=False)
+
+    # File to process
+    input_filepath = Column(String(1024), nullable=False, index=True)
+
+    # generate_climos parameters
+    output_directory = Column(String(1024), nullable=False)
+    convert_longitude = Column(Boolean, default=True, nullable=False)
```
jameshiebert: Again, smart to keep track of these parameters and preserve configurability through the whole stack.

dp/jobqueueing/alembic/env.py
```diff
@@ -0,0 +1,105 @@
...
+
+The `<db-name>` section must appear in `alembic.ini`, with a `sqlalchemy.url =` line.
+For example, to use `alembic -x db=test upgrade ...`::
+
+    [test]
+    sqlalchemy.url = sqlite:////path/to/database/test.sqlite
+
+Adapted from https://gist.github.com/twolfson/4bc5813b022178bd7034
+"""
+from __future__ import with_statement
```
rod-glover: This is generic code that comes with Alembic. We can modify it, but I usually prefer to leave what amounts to the source of an imported package alone.

dp/jobqueueing/gcsub.py
rod-glover: Yeah, some of this was done in haste -- "fix it later" mode. I'll see to that.

dp/jobqueueing/gcsub.py
rod-glover: My thought was that we'd establish a standard environment independently of the scripts that use it. But you are right that it is fragile.

The naive approach would be to set up a new environment for every run of the script. It could even include a git pull on the repo to ensure that we are running the latest greatest. But it has the following issues to be addressed:

- Concurrency: These jobs only require one processor each, so typically 4 run at a time. If each script is running setup, they will interfere with each other. They'll have to establish their own uniquely named virtual env to run in. Alternatively we clone the whole repo to a uniquely named directory.
- Speed: There will be overhead for setting up each time. Not sure how much.
- Trickiness, time to dev: I recall that setting up the environment took some hacking about each time. Possibly due to my own shortcomings ... but a clean, error-free setup script has to be established and that will also take some time.

dp/jobqueueing/gcupd-qstat.py
rod-glover: Yeah, quickest shortcut to a parser for that output that I could come up with.

dp/jobqueueing/alembic/env.py
rod-glover: OTOH, I already modified it to handle the database configuration ...

dp/jobqueueing/gcsub.py
jameshiebert: > My thought was that we'd establish a standard environment independently of the scripts that use it.

Yeah, I was thinking the same thing.

dp/jobqueueing/gcreset.py
rod-glover: Apologies. This was developed pretty quickly, and I actually changed my mind about the workflow partway through (from emails to output of qstat). I'll make some major improvements to the documentation.

dp/jobqueueing/gcalter-params.py
rod-glover: Yeah, given that I wanted to get this working fairly quickly, I decided not to risk getting bogged down in some SQLite rabbit hole (relatively low risk, I now think, but ...).

dp/jobqueueing/gcsub.py
rod-glover: Good. And it seems to me that the virtual environment is part of that setup. Is there some way you would like to abstract away from the specific path?

rod-glover: I do think that the standard environment (repo clone) should not be in my home directory. I'll move it to somewhere stable, e.g., `/storage/data/projects/comp_support/climate_exporer_data_prep/climatological_means/`

dp/jobqueueing/gcsub.py
rod-glover: See comment above

rod-glover:
Python virtual env is now stored in the database (per queue entry), and is a required parameter of the `gcadd` script. That parameter defaults to the value of the environment variable `GCQ_PY_ENV` (in the environment in which `gcadd` is invoked, not the `generate_climos` execution environment), if it exists. It is required either from the env or on the command line. Command line overrides.

Database migration defaults the value for this column in existing records to `/storage/data/projects/comp_support/climate_exporer_data_prep/climatological_means/venv`, which is exactly where I intend to install the relevant script(s).

Default jobqueue database is now defaulted to the value of environment variable `GCQ_DATABASE`. Ditto all comments with respect to `GCQ_PY_ENV`.

rod-glover:
Ready for a second round of review.

rod-glover:
Jobqueueing has been extracted to a separate project repo (https://github.com/pacificclimate/jobqueueing). Therefore this PR is being closed without merging.

PR #51, opened 2017-06-30
Fix boolean CLI arguments

rod-glover:
Fix incorrect parsing of boolean arguments -g, -v, -i.


corviday:
Looks pretty good to me.

jameshiebert:
Seems reasonable.

PR #60, opened 2017-07-25
Handle climo files with separate chronology

### LINKED ISSUE #57: Data API call doesn't handle split climatology files
The [data API call](https://github.com/pacificclimate/climate-explorer-backend/blob/master/ce/api/data.py) collects and returns data from all files that match a set of parameters (model, experiment, ensemble, variable, etc). 

Previously, it took a `time` index argument between 0 and 16 to indicate time resolution and position sought, and would return the nth time slice in the file, which worked when all the files were 17-point chronologies.

Using split data files, with separate monthly (0-11), seasonal (0-3) and yearly(0) files, this API call returns a mix of monthly, seasonal, and annual data when you call it with `&time=0`, and an `IndexError` if you call it with any other time index.

Here's the result for `GET /api/data?ensemble_name=downscaled&model=bcc-csm1-1-m&variable=tasmax&emission=historical,+rcp45&area=&time=0`:
```JSON
{
  "r1i1p1": {
    "units": "degC",
    "data": {
      "2085-07-02T00:00:00Z": -0.5808153910727394,
      "2025-07-02T00:00:00Z": -0.8622292459716103,
      "1977-01-16T00:00:00Z": -18.78877913696885,
      "1997-01-16T00:00:00Z": -17.92588300616977,
      "2055-07-02T00:00:00Z": -0.6485861922775469,
      "2055-01-16T00:00:00Z": -15.875643052269846,
      "1986-07-02T00:00:00Z": -2.4333102893767307,
      "1997-07-02T00:00:00Z": -2.077599634237314,
      "2025-01-16T00:00:00Z": -16.226023053533392,
      "2025-01-15T00:00:00Z": -18.02569051912951,
      "2085-01-16T00:00:00Z": -15.60062807695439,
      "1986-01-15T00:00:00Z": -19.950983475060585,
      "1977-07-02T00:00:00Z": -2.671051067797724,
      "1977-01-15T00:00:00Z": -20.599000150601793,
      "2055-01-15T00:00:00Z": -17.825752320828578,
      "1986-01-16T00:00:00Z": -18.487454919078537,
      "1997-01-15T00:00:00Z": -19.534196834187902,
      "2085-01-15T00:00:00Z": -17.498223073165622
    }
  }
}
```

This appears to be a mix of annual data (July 2 dates), monthly data (January 15), and seasonal data (January 16).
screenshot from 2017-07-19 15-49-44

I don't have strong preferences about how exactly this query should behave in the split-file context. It doesn't need to function identically to the old query as long as it's possible to get the data in some reasonably straightforward way. For example, if given the new file structure it makes sense to expect the front end to pass a time resolution (monthly, seasonal, annual) along with the time index, that would be no trouble.

rod-glover:
Fixes #57 

The fix itself was simple. Modifying the tests to replace completely non standards-compliant test files `cgcm.nc` and `cgcm-tmin.nc` with standards compliant ones took most of the work.


ce/api/data.py
```diff
@@ -72,40 +72,79 @@ def data(sesh, model, emission, time, area, variable, timescale='other',
...
+        .filter(Model.short_name == model)
+
+        .join(Run.emission)
+        .filter(Emission.short_name == emission)
+
+        .join(DataFile.timeset)
+        .filter(TimeSet.time_resolution == timescale)
+
+        .filter(DataFileVariable.ensembles.any(Ensemble.name == ensemble_name))
+    )
```
rod-glover: I grouped related joins and filters, but this may not seem to others, as it does to me, a clearer way to organize the query.

corviday:
```
     Searches the database for all files from a given model and
     emission scenario and returns a data value for that particular
     timestep (e.g. January [1], summer [14] or annual [17]) in each
     file.
```

Maybe I'm misunderstanding how this query works now, but I think this docstring in `data.py` would benefit from slight revision. You'd now pass 0 for January (with `timescale=monthly`), 2 for summer (with `timescale=seasonal`) and 0 for annual (with `timescale=yearly`)? Am I understanding that correctly?

corviday:
Seems to produce plausible data and accept the specified inputs. 
screenshot from 2017-07-25 17-00-18

ce/tests/test_api.py
```diff
@@ -168,28 +183,67 @@ def test_data_bad_time(populateddb):
...
+@pytest.mark.parametrize('variable', (
+        'tasmax',
+        'tasmin',
+))
+@pytest.mark.parametrize('timescale, time_idx, expected_ymd', (
+        ('monthly', 8, (1985, 9, 15)),
+        ('seasonal', 2, (1985, 7, 15)),
+        ('yearly', 0, (1985, 7, 2)),
+))
+def test_data_single_file(populateddb, variable,
```
jameshiebert: Do we do any exception testing here? E.g. what happens if you give `time_idx=1` for `timescale='yearly'`, etc.?

corviday: [Here's what happens at present](http://docker2.pcic.uvic.ca:8300/api/data?ensemble_name=downscaled&model=bcc-csm1-1-m&variable=tasmax&emission=historical,+rcp45&timescale=monthly&area=&time=12). 

The frontend data validation functions [check to see](https://github.com/pacificclimate/climate-explorer-frontend/blob/master/src/components/DataControllerMixin/DataControllerMixin.js#L93) if the backend returns a string object instead of a JSON object to catch that sort of thing.

There may be a better way to deal with this, but that's what happens now.

jameshiebert: Yikes, yeah we definitely want to avoid the backend returning 500s!

PR #62, opened 2017-08-09
Fix #59 - multimeta API exposes climatology bounds

### LINKED ISSUE #59: Multimeta API call should expose time bounds
The single-file [metadata API call](https://github.com/pacificclimate/climate-explorer-backend/blob/master/ce/api/metadata.py) now includes `start_date`, `end_date`, and `multi_year_mean` metadata about each dataset, as requested in #44 . It would be convenient if the multi-file [multimeta API call](https://github.com/pacificclimate/climate-explorer-backend/blob/master/ce/api/multimeta.py) exposed them too, but I'm open to discussion about the necessity of this.

At present, the climate explorer frontend [calls `multimeta` once on loading (line 17)](https://github.com/pacificclimate/climate-explorer-frontend/blob/master/src/components/AppMixin/AppMixin.js) to get information on all available datasets, which it uses to populate selectors and user-facing widgets with what datasets are available for examination. 

In order to be able to show the user what each dataset's climatological period is, either `multimeta` could provide that information, or climate-explorer-frontend could hit the single-file `metadata` call once for each file in the user-selected current group (ie, filtered by model, variable, and sometimes experiment). I think having `multimeta` provide the information up front would be a little nicer.

corviday:


jameshiebert:
It would probably be worth adding a couple of tests (or checks to existing tests). Especially given that there is conditional behaviour here. So you probably want at least one test that checks that the result *does* have start/end date and one test that checks it does *not*.

corviday:
Good idea, thanks.

corviday:
I'd incorrectly thought that datasets that weren't multi year means didn't have defined `start_date` and `end_date` since they don't correspond to a climatology. They actually do have `start_date` and `end_date`:  the earliest and latest timestamps with data associated. 

So this query is a lot simpler than I'd been making it. No conditional needed, just returns` start_date`. `end_date`, and `multi_year_mean` for all datasets. Accordingly, only the most insignificant additions to the test suite were needed.

corviday:
Travis tests are failing due to the recent update of [Shapely](https://pypi.python.org/pypi/Shapely) from version 1.5 to 1.6. The update changed the location of `ReadingError`, which is used in [test_geo.py](https://github.com/pacificclimate/climate-explorer-backend/blob/master/ce/tests/test_geo.py). PR https://github.com/pacificclimate/climate-explorer-backend/pull/68 fixes that issues and should be merged first.

jameshiebert:
This looks great. Feel free to merge once it's rebased onto #68

Issue #65, opened 2017-08-21
Flat data from timeseries API call when called with a spatial extent


corviday:
The [timeseries API call](https://github.com/pacificclimate/climate-explorer-backend/blob/master/ce/api/timeseries.py), when passed a spatial extent to calculate values over, returns [values that do not vary over the year](http://docker2.pcic.uvic.ca:8601/api/timeseries?id_=tasmin_mClim_BCCAQv2_CanESM2_historical-rcp26_r1i1p1_19610101-19901231_Canada&variable=tasmin&area=POLYGON+((-129.4677734375+50.809326171875,+-128.63769531250003+51.30981445312501,+-126.67236328125+50.797119140625,+-124.37744140624999+49.942626953125,+-123.095703125+48.917236328125,+-122.83935546875001+48.160400390625,+-124.8046875+48.13598632812501,+-128.173828125+49.55200195312501,+-129.4677734375+50.809326171875))):
```JSON
{
  "id": "tasmin_mClim_BCCAQv2_CanESM2_historical-rcp26_r1i1p1_19610101-19901231_Canada",
  "units": "degC",
  "data": {
    "1977-01-15T00:00:00Z": 4.026229503566938,
    "1977-02-15T00:00:00Z": 4.026229503566938,
    "1977-03-15T00:00:00Z": 4.026229503566938,
    "1977-04-15T00:00:00Z": 4.026229503566938,
    "1977-05-15T00:00:00Z": 4.026229503566938,
    "1977-06-15T00:00:00Z": 4.026229503566938,
    "1977-07-15T00:00:00Z": 4.026229503566938,
    "1977-08-15T00:00:00Z": 4.026229503566938,
    "1977-09-15T00:00:00Z": 4.026229503566938,
    "1977-10-15T00:00:00Z": 4.026229503566938,
    "1977-11-15T00:00:00Z": 4.026229503566938,
    "1977-12-15T00:00:00Z": 4.026229503566938
  }
}
```
The same query run over the entire spatial extent returns[ expected seasonal variation](http://docker2.pcic.uvic.ca:8601/api/timeseries?id_=tasmin_mClim_BCCAQv2_CanESM2_historical-rcp26_r1i1p1_19610101-19901231_Canada&variable=tasmin&area=):
```JSON
{
  "id": "tasmin_mClim_BCCAQv2_CanESM2_historical-rcp26_r1i1p1_19610101-19901231_Canada",
  "units": "degC",
  "data": {
[...]

PR #66, opened 2017-08-25
Slice area-masked array by time index

corviday:
The issues in #65 (stats and timeseries API results not varying over the course of the year) seem to have been caused by `util.get_array()` returning data for all time indexes at once in a 3D array [time:lat:lon] when passed an area. Then `numpy.mean()` would calculate the average of the _entire_ array - all four or twelve time indexes, resulting in a "timeseries" of twelve/4 identical values, each of which was the mean of the entire year instead of a single month / season. When no area was passed, a 2D array [lat:lon] would be returned.

I remain uncertain, however, because this seems to have been working at some point in the past - at least in one specific docker container whose functionality I cannot duplicate (even by downloading and running the same image from docker hub) - and there's no code change that would explain why it stopped working. Makes me suspicious that something else is going on.

So I would definitely appreciate eyes on this fix.

### DIFF
--- a/ce/api/util.py
+++ b/ce/api/util.py
@@ -64,19 +64,34 @@ def get_array(nc, fname, time, area, variable):
 
     a = nc.variables[variable]
 
-    # FIXME: Assumes 3d data... doesn't support levels
-    if time or time == 0:
-        assert 'time' in nc.variables[variable].dimensions
-        a = a[time,:,:]
-    else:
-        a = a[:,:,:]
-
     if area:
         # Mask out data that isn't inside the input polygon
         a = wkt_to_masked_array(nc, fname, area, variable)
+        a = time_slice_array(a, time, nc, fname, variable)
     else:
+        a = time_slice_array(a, time, nc, fname, variable)
         a = ma.masked_array(a)
 
+    return a
+
+# FIXME: Assumes 3d data... doesn't support levels
+#Reduces a 3-dimensional array to a two-dimensional array by
+#returning the timeidxth slice, IFF time is defined and time
+#is a dimension in the netCDF. Otherwise return array unchanged.
+def time_slice_array(a, timeidx, nc, fname, variable):
+    if timeidx or timeidx == 0:
+        if 'time' not in nc.variables[variable].dimensions:
+            raise Exception(
+                "File {} does not have a time dimension".format(fname)
+                )
+        if not isinstance(timeidx, int): #times passed by URL may be strings
+            try:
+                timeidx = int(timeidx)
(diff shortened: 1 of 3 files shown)


corviday:
[Demo timeseries query](http://docker2.pcic.uvic.ca:8703/api/timeseries?id_=tasmin_mClim_BCCAQv2_bcc-csm1-1-m_historical-rcp45_r1i1p1_19610101-19901231_Canada&variable=tasmin&area=POLYGON+((-126.33056640625001+50.9649658203125,+-126.91650390625001+50.0250244140625,+-125.82397460937501+49.4635009765625,+-124.951171875+50.6597900390625,+-126.32446289062499+50.9344482421875,+-126.33056640625001+50.9649658203125))) for this branch.

corviday:
Since there are several pull requests failing tests on this repository, I split the fix to the `shapely.ReadingError` location - originally part of this PR - into its own pull request, https://github.com/pacificclimate/climate-explorer-backend/pull/68 . That should be merged first, then this branch rebased, reviewed,  and merged.

ce/api/util.py
```diff
@@ -64,21 +64,36 @@ def get_array(nc, fname, time, area, variable):
...
+#Reduces a 3-dimensional array to a two-dimensional array by
+#returning the timeidxth slice, IFF time is defined and time
+#is a dimension in the netCDF. Otherwise return array unchanged.
+def time_slice_array(a, timeidx, nc, fname, variable):
+    if timeidx or timeidx == 0:
+        if 'time' not in nc.variables[variable].dimensions:
+            raise Exception(
+                "File {} does not have a time dimension".format(fname)
+                )
+        if not isinstance(timeidx, int): #times passed by URL may be strings
```
jameshiebert: I would suggest putting these types of checks farther up the call stack. I.e. if an HTTP response is expected to send a valid ``int``, make those checks as soon as you process the query string. At this level, we should be able to assume that we're getting an ``int``.

jameshiebert:
So, I have reviewed this independently by stepping through the code during a call to ``get_array()``. This was an excellent catch on your part. The problem is very clearly the fact that previously the temporal subset operation was done *before* the call to ``wkt_to_masked_array()`` and ``wkt_to_masked_array()`` didn't actually take an array as a parameter (it grabs the data straight out of the NetCDF file).

If we wanted to avoid some extra computation, we could rewrite to do the temporal subset first and *then* do the spatial subset and scaling. However, we're probably dealing with so little data that it won't matter.

As far as I can tell, your fix here is exactly what we need to do.

PR #68, opened 2017-09-08
Update location of ReadingError

corviday:
Shapely's update from version 1.5 to version 1.6 on August 20th included consolidating all the package's errors in one file. One of the tests in `test_geo.py` references `shapely.ReadingError` and needs to be updated with the new error location.


jameshiebert:
The only thing that I would (strongly) suggest here, is that you update requirements (in `setup.py` *and* `requirements.txt`) to require Shapely >= 1.6.

jameshiebert:
W00t!

Issue #70, opened 2017-11-02
Area query on climdex file causes server crash


corviday:
[Timeseries query](https://github.com/pacificclimate/climate-explorer-backend/blob/master/ce/api/timeseries.py) and [multistats query](https://github.com/pacificclimate/climate-explorer-backend/blob/master/ce/api/multistats.py) crash the entire backend when made against a climdex file with an area specified.

corviday:
My first suspicion was that this was another manifestation of [this data prep issue](https://github.com/pacificclimate/climate-explorer-data-prep/issues/11), but the crash occurs with files that have monotonic longitudes as well.

corviday:
The crash occurs inside wrappered GDAL routines called with what appear to be entirely valid arguments. I poked around with GDB , but wasn't able to sort out anything useful or solve the issue from the GDAL end.

Instead I upgraded rasterio to the bleeding edge alpha version, because I know rasterio changed some of how transforms work with GDAL in the newest versions and I thought it might be different enough to avoid whatever mysterious thing was causing GDAL to crash. That seems to have fixed the issue. Need to do a little more testing before I'm sanguine about switching to the alpha version of such an essential library.

jameshiebert:
OK, I'm in the process of upgrading our [geospatial-python docker image](https://hub.docker.com/r/pcic/geospatial-python/builds/) to use GDAL==2.2. Do you suspect that using that will help?

corviday:
Seems like it would probably help.

jameshiebert:
Example request that fails:

http://docker-dev01.pcic.uvic.ca:30123/api/timeseries?id_=altwsdiETCCDI_yr_BCCAQ_CNRM-CM5_historical-rcp85_r1i1p1_19500702-21000702&variable=altwsdiETCCDI&area=POLYGON+((-133.8134765625+51.14501953125,+-133.8134765625+54.90478515625,+-130.0048828125+54.90478515625,+-130.0048828125+51.14501953125,+-133.8134765625+51.14501953125))

or:

http://[host]:[port]/api/timeseries?id_=altwsdiETCCDI_yr_BCCAQ_CNRM-CM5_historical-rcp85_r1i1p1_19500702-21000702&variable=altwsdiETCCDI&area=POLYGON+((-133.8134765625+51.14501953125,+-133.8134765625+54.90478515625,+-130.0048828125+54.90478515625,+-130.0048828125+51.14501953125,+-133.8134765625+51.14501953125))

[...]

PR #71, opened 2017-11-16
Fix GDAL crash

corviday:
This branch is the result of efforts to fix a crash in gdal. I wasn't initially sure of the cause of the bug and worked on several possible theories, so there are multiple changes in this branch. One of them (the rasterio version update) fixed the original bug. Two other changes seemed like they improved the code in worthwhile ways even though as far as fixing the gdal bug they turned out to be dead ends, so I have left them in the branch.

Changes in this branch:
* Switched versions of rasterio; the newest versions use different GDAL wrappers and do not trigger bug #70 
* ~~Replaced the [shapely](https://github.com/Toblerity/Shapely) module with the [geomet](https://github.com/geomet/geomet) module for wkt processing, following the rasterio documentation at [Switching From GDAL's Python Bindings](https://mapbox.github.io/rasterio/switch.html). I was hoping this would solve a crash that occurs when the backend is run against GDAL version 2.2 in a docker. While this did not fix the problem, given rasterio's recommendations, I left the module switch in place to forestall possible future problems.~~
 * Updated caching - made cache maxsize a decorator argument, increased cache size to 100MB, fixed keys for caching masked data, added caching of grid masks in addition to caching masked data

Question for reviewers of this branch:
* I haven't written python decorators before (they're cool! that was fun!). Would someone look over `@memoize` and make sure it's readable and not unnecessarily complicated?

**EDIT**: Shapely removal has been reverted.


ce/api/geo.py
```diff
@@ -52,88 +53,144 @@ def inner(obj, _seen_ids = None):
 log = logging.getLogger(__name__)
 cache_lock = RLock()
 
-class memoize_mask(object):
+class memoize(object):
```
jameshiebert: Not saying that you necessarily *should* do this (or that I'm requesting a change to this PR), but I believe that this decorator could be written more elegantly in a functional style than in an OO style. For example, I think that all of the state that you store in `self.*` in the `__init__()` functions could just as easily come from a closure context.

But this works too (in a slightly more verbose manner); just want you to be aware of the various stylistic options.

ce/api/geo.py
```diff
@@ -52,88 +53,144 @@ def inner(obj, _seen_ids = None):
...
     '''
-    Decorator. Caches wktToMask keyed to model_id and the WKT string
+    Decorator to add caching to a function.
+    Initialized with a key-generating function and maxsize. Caches up to maxsize
+    MB worth of results from the decorated function using a key created from the
+    decorated function's arguments by the key function.
+    A key generator should have the same signature as the cached function.
     '''
-    def __init__(self, func, maxsize=50):
+    def __init__(self, keyfunc, maxsize=50):
```
jameshiebert: Glad that this maxsize parameter *actually* works now :)

ce/api/geo.py
```diff
@@ -52,88 +53,144 @@ def inner(obj, _seen_ids = None):
...
+
+    return mask, out_transform, window
 
-        the_array, _ = rio_mask(raster, [mapping(poly)], crop=False, all_touched=True)
+def make_masked_file_key(nc, fname, wkt, varname):
+    """generates a key suitable for characterizing a masked netCDF file:
+       filename and polygon"""
+    return (fname, wkt)
+
+@memoize(make_masked_file_key, 100)
```
jameshiebert: So I would propose that we make the maxsize argument here parameterizable as an application option. This is something that we'll probably want to change based on the deployment. 100MB might be too much for working on our laptops, but we could probably afford more on  larger deploy with dedicated resources. Having some flexibility available from the top level would be nice. Let's add that as a feature request... doesn't need to be part of this PR.

ce/api/geo.py
```diff
@@ -52,88 +53,144 @@ def inner(obj, _seen_ids = None):
...
 
+            def get_size(self): return getsize(self.cache)
 
-def polygon_to_masked_array(nc, fname, poly, variable):
+            def get_length(self): return len(self.cache)
+
+        cf = CachedFunction(self.maxsize, self.keyfunc, func)
+        return cf
 
+def make_mask_grid_key(nc, fname, poly, variable):
```
jameshiebert: Yeah, this is kind of a pain to pull all of the grid information out, so good job on stashing it all away in it's own function.

ce/api/geo.py
```diff
@@ -52,88 +53,144 @@ def inner(obj, _seen_ids = None):
...
+    latsteps = nc.variables['lat'].shape[0]
+    latmin = nc.variables['lat'][0]
+    latmax = nc.variables['lat'][latsteps - 1]
+    lonsteps = nc.variables['lon'].shape[0]
+    lonmin = nc.variables['lon'][0]
+    lonmax = nc.variables['lon'][lonsteps - 1]
+    wkt = dumps(poly) #dict-style geoJSON can't be hashed
+    return (wkt, latmin, latmax, latsteps, lonmin, lonmax, lonsteps)
+
+@memoize(make_mask_grid_key, 10)
```
jameshiebert: Perfect. We should get a lot more cache hits with this and avoid the actually expensive operation.

ce/api/timeseries.py
```diff
@@ -59,13 +59,14 @@ def timeseries(sesh, id_, area, variable):
...
     ti.sort(key=lambda x: x[1])
-
+    
     with open_nc(file_.filename) as nc:
 
         data = OrderedDict([(
             timeval.strftime('%Y-%m-%dT%H:%M:%SZ'),
             np.asscalar(np.mean(get_array(nc, file_.filename, idx, area, variable)))
         )
+            #for timeval, idx in shortlist
```
jameshiebert: Should have cut this line.

requirements.txt
```diff
@@ -5,9 +5,8 @@ modelmeta
 shapely>=1.6
 numpy
 netcdf4
-nchelpers
 GDAL
-rasterio
+rasterio==1.0a11
```
jameshiebert: We'll want to remember to upgrade this when 1.0 is out of alpha :)

PR #72, opened 2017-11-17
Use Flask's command line interface

jameshiebert:
The Flask docs state that starting with version 0.11 one *should* use the flask CLI, rather than starting Flask from inside your Python modules.
http://flask.pocoo.org/docs/0.12/server/#in-code says:
"The reason for this is that due to how the reload mechanism works
there are some bizarre side-effects (like executing certain code
twice, sometimes crashing without message or dying when a syntax or
import error happens)."
Following this advice, this PR removes the "devserver.py" start up script and replaces it with guidance on how to start the application directly with Flask's CLI.


README.md
```diff
@@ -37,7 +37,7 @@ $ source venv/bin/activate
 
 Database dsn can be configured with the MDDB_DSN environment variable. Defaults to 'postgresql://httpd_meta@monsoon.pcic.uvic.ca/pcic_meta'
 
-(venv)$ MDDB_DSN=postgresql://dbuser:dbpass@dbhost/dbname python scripts/devserver.py -p <port>
+(venv)$ MDDB_DSN=postgresql://dbuser:dbpass@dbhost/dbname FLASK_APP=ce.wsgi:app flask run -p <port>
```
corviday: Maybe mention that one could do FLASK_DEBUG=1 here to enable live reloading on code change?

jameshiebert: Yeah, good suggestion.

PR #75, opened 2018-02-07
Set HTTP headers to enable client-side caching

### LINKED ISSUE #74: Provide HTTP cache support
There are numerous methods in HTTP to provide support for caching:

1. [`Last-Modified` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified)
1. [`ETag`s](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag)
1. [max-age](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching)

Our data changes almost never, so we could easily provide some very aggressive and particularly accurate caching control, using the mod times of the underlying data files. If we do this, we could provide much better performance and reduce the system load by a ton.

jameshiebert:
This PR fixes #74 by setting two HTTP caching headers, allowing clients to forgo the need to refresh rarely changing content.

## Cache-Control

We set `cache-control` to "public", essentially turning it on, and we set a `max-age` of 24 hours (climate data will rarely-if-never be updated more often than that).

## Last-Modified

For requests that are tied to one or more ``DataFile`` entries, we set the last-modified header to be the most recent update seen on any of the data files that comprise the response. The update times are taken from the `index_time` field of the ``DataFile`` entries.

## Byproducts

To implement this, I ended up changing the internal representation of date times. Rather than using ``strftime()`` to format them when they come right out of the database, I opted to use the Python ``datetime`` object to pass values through the code, only serializing right before its necessary to encode to JSON. I think this makes the code a bit cleaner (and it was necessary for testing).

## Demo

Backend demo is [here](http://docker-dev01.pcic.uvic.ca:30888/api).
Frontend demo using this backend is [here](http://docker-dev01.pcic.uvic.ca:30889). Watch your dev tool's network requests and note that it shouldn't make a re-request when you switch back-and-forth between two options.

The demo isn't using the production Gunicorn work pool for servicing requests, so it *may* appear a bit sluggish if requests back up behind each other. But just pay attention to what your browser is requesting and more importantly, what it isn't.

Comments/questions/concerns welcome.


ce/api/metadata.py
```diff
@@ -50,23 +50,22 @@ def metadata(sesh, model_id):
...
+                        1: datetime.datetime(1985, 2, 15, 0, 0),
+                        2: datetime.datetime(1985, 3, 15, 0, 0),
+                        3: datetime.datetime(1985, 4, 15, 0, 0),
                     }
                     'multi_year_mean': False,
-                    'start_date': '1985-01-15T00:00:00Z',
-                    'end_date': '1985-04-15T00:00:00Z',
+                    'start_date': 'datetime.datetime(1985, 1, 15, 0, 0),
+                    'end_date': datetime.datetime(1985, 4, 15, 0, 0),
+                    'modtime': datetime.datetime(2010, 1, 1, 17, 30, 4)
```
rod-glover: I really approve of keeping datetimes in native representation until final conversion.

ce/tests/test_api.py
```diff
@@ -32,6 +35,23 @@ def test_api_endpoints_are_callable(test_client, cleandb, endpoint, query_params
     url = '/api/' + endpoint
     response = test_client.get(url, query_string=query_params)
     assert response.status_code == 200
+    assert response.cache_control.public == True
+    assert response.cache_control.max_age > 0
```
rod-glover: Do you want to be a bit more specific about `max_age`?

ce/tests/test_api.py
```diff
@@ -77,23 +97,24 @@ def test_metadata(populateddb, unique_id):
...
                 'variables', 'ensemble_member', 'times', 'timescale',
-                'multi_year_mean', 'start_date', 'end_date']:
+                'multi_year_mean', 'start_date', 'end_date', 'modtime']:
         assert key in file_metadata
 
     times = file_metadata['times']
     assert len(times) > 0
 
-    # Are the values converible into times?
+    # Are the values propert datetimes?
```
rod-glover: "proper"?

ce/tests/test_api.py
```diff
@@ -119,6 +140,7 @@ def test_multimeta(populateddb, model):
     # make sure start_date and end_date are present
     assert 'start_date' in rv[unique_id]
     assert 'end_date' in rv[unique_id]
+    assert 'modtime' in rv[unique_id]
```
rod-glover: Should you be asserting the types (datetime) of these?

ce/tests/test_api.py
```diff
@@ -176,6 +199,7 @@ def test_stats_bad_params(populateddb, unique_id, var):
     assert math.isnan(rv[unique_id]['max'])
     assert 'time' not in rv[unique_id]
     assert 'units' not in rv[unique_id]
+    assert 'modtime' not in rv[unique_id]
```
rod-glover: Assert types (datetime) as well?

rod-glover:
Looks good. I think the switch to a native representation for dates until response is generated is a good idea in any case. 

A couple of minor issues, mostly to do with testing.

ce/tests/test_api.py
jameshiebert: Meh. We only set it in one and only one place, and the resolution of caching age is so coarse as to be trivial. I'm really only concerned that it got properly set to be anything aside from zero.

ce/tests/test_api.py
jameshiebert: Yep, good catch.

ce/tests/test_api.py
jameshiebert: Yes, good suggestion.

ce/tests/test_api.py
jameshiebert: This is a ``not in`` check, so there's no type to assert.

PR #80, opened 2018-06-21
Upgrade rasterio to new release candidate (from alpha)

jameshiebert:
All of the tests pass (*really* quickly) on my desktop. Is there any reason *not* to do this? I don't seem much in [CHANGES.txt](https://github.com/mapbox/rasterio/blob/master/CHANGES.txt) that should affect us, but there are a few mentions of `rasterize()`, so maybe it has seen some underlying work.


corviday:
I've been testing with it too, though I'm only using 1.0b1 and I have had no issues (related to rasterio) either. I see no reason not to go ahead once tests are passing. 

Tests are failing due to a netcdf4 update and a flask update. I ended up just nailing netcdf4 at 1.3.1 and switching every instance of `flask.ext.sqlalchemy` to `flask_sqlalchemy`.

PR #82, opened 2018-07-09
Use only multi year means for /data query

corviday:
The` /data` API endpoint is called by the code that generates the Long Term Average graph if the user has selected a multi year mean dataset; it yields a series of values, one for each available period, illustrating change over time. 

This query also adds non-mym datasets to the series, which is not correct, but wasn't an issue and didn't come up until we started using an ensemble with both nominal-time and MYM data.


jameshiebert:
That's a quick and easy fix. Thanks for finding that!

PR #84, opened 2018-07-16
Add tests for cache deletion

corviday:
Minor change that adds a test to make sure the cache deletes items when it is full. We don't have (and probably don't want to add) enough test data netCDFs to fill up the 100MB data cache, so the new test uses a recursive Fibonacci function to test a 10-int-sized cache.

The failed test is an unrelated timing test, which just barely wasn't fast enough. I don't think it represents a genuine issue.

### DIFF
--- a/ce/tests/test_geo.py
+++ b/ce/tests/test_geo.py
@@ -1,10 +1,11 @@
 import time
-
 import pytest
 
-from ce.api.geo import wkt_to_masked_array, polygon_to_masked_array, polygon_to_mask
+from ce.api.geo import wkt_to_masked_array, polygon_to_masked_array
+from ce.api.geo import polygon_to_mask, memoize, getsize
 from shapely.wkt import loads
 from shapely.errors import ReadingError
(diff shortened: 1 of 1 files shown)


jameshiebert:
I know that it's janky, but you can often re-run the tests and get the timing tests to pass (we should probably pytest.mark the timing tests and turn them off on Travis).

ce/tests/test_geo.py
```diff
@@ -51,6 +51,44 @@ def test_mask_cache(netcdf_file):
...
+    f = cached_fibonacci
+    
+    previous_cache = f.get_size()
+    cache_shrunk = False
+    
+    for n in range(1, 20):
+        cached_fibonacci(n)
+        assert (f.get_size() / mb_conversion) <= cache_size_mb
+        if(not cache_shrunk and previous_cache <= f.get_size()):
+            cache_shrunk = True
```
jameshiebert: At this point, couldn't you just return `True` as pass the test?

PR #87, opened 2018-08-10
Initial design for the streamflow RESTful API

corviday:
@rod-glover, does this seem like a good place to store the evolving API design?


corviday:
Rod is working on a different branch that adds the swagger file.

PR #88, opened 2018-08-14
Add streamflow API spec

### LINKED ISSUE #86: Add streamflow API
1. Define streamflow API in OpenAPI/Swagger
2. Implement API
   * This will require accompanying changes to the modelmeta database.

rod-glover:
Resolves #86 

Updated from initial draft circulated a few weeks ago. Updates:

- Base for streamflow-related endpoints: `services.pacificclimate.org/streamflow/` (not routed_streamflows).
- All resource names plural (e.g., `/streamflow/orders`, not `/streamflow/order`).
- All response bodies are fully fleshed out (or should be).
- Consistent use of hyperlinks in all response bodies.
- Hyperlink `ref` properties are abbreviated (e.g., 'cancel') instead of long-form hyperlink.
- Full factoring of commonalities in schemas, parameters, responses, etc.

Notes: 

- The [Streamflow Architecture document](https://pcic.uvic.ca/confluence/display/CSG/Streamflow+Architecture+II) is a key resource to understanding the lifecycles of Orders and Results. 
- See also [Streamflow API Design discussion](https://pcic.uvic.ca/confluence/display/CSG/Streamflow+API+Design).

### DIFF
--- a/streamflow-api.yaml
+++ b/streamflow-api.yaml
@@ -78,6 +78,23 @@ paths:
         404:
           $ref: '#/components/responses/404NotFound'
 
+  /hydromodel_outputs/{id}/bounds:
+    get:
+      summary: Get bounds of a specific VIC hydrological model output.
+      tags:
+        - Hydromodel outputs
+      parameters:
+        - $ref: '#/components/parameters/hydromodelOutputId'
+      responses:
+        200:
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HydromodelOutputBounds'
+        404:
+          $ref: '#/components/responses/404NotFound'
+
   ### Orders
 
   /streamflow/orders:
@@ -451,12 +468,20 @@ components:
             experiment:
               type: string
               example: rcp85
+            cell_x_size:
+              type: number
+              example: 3.75
+              description: Size of grid cells in X (longitude) direction.
+            cell_y_size:
+              type: number
+              example: 2.5
+              description: Size of grid cells in Y (latitude) direction.
             etc:
               type: string
               example: ... more tbd ...
 
     HydromodelOutputHypermedia:
-      description: Hypermedia for a thing
+      description: Hypermedia for a hydromodel output.
       type: object
       properties:
         links:
@@ -465,12 +490,43 @@ components:
             - example:
                 - rel: 'self'
                   uri: 'http://apis.paccifcclimate.org/hydromodel_output/4'
+                - rel: 'bounds'
+                  uri: 'http://apis.paccifcclimate.org/hydromodel_output/4/bounds'
 
     HydromodelOutput:
       description: Full representation of a hydromodel output - state + hypermedia.
       allOf:
         - $ref: '#/components/schemas/HydromodelOutputState'
         - $ref: '#/components/schemas/HydromodelOutputHypermedia'
+
+    HydromodelOutputBounds:
+      description: Geographic bounds of hydromodel data set.
+      type: object
+      properties:
+        bounds:
+          type: object
+          description: GeoJSON defining the bounds
+        links:
+          - $ref: '#/components/schemas/HypermediaLinkList'
+      example:
+        bounds:
+          type: Feature
+          properties:
+            name: Columbia Basin
+            area: 1000000
+          geometry:
+            type: Polygon
+            coordinates:
+              -
+                - [-64.73, 32.31]
+                - [-80.19, 25.76]
+                - [-66.09, 18.43]
+                - [-64.73, 32.31]
+        links:
+          - rel: 'self'
+            uri: 'http://apis.paccifcclimate.org/hydromodel_output/4/bounds'
+          - rel: 'hydromodel_output'
+            uri: 'http://apis.paccifcclimate.org/hydromodel_output/4'
 
     # Lists (collections) of hydromodel outputs
 
--- a/streamflow-api.yaml
+++ b/streamflow-api.yaml
@@ -252,6 +252,36 @@ paths:
         404:
           $ref: '#/components/responses/404NotFound'
 
+  /streamflow/results/{id}/annual-cycle/standard-periods:
+    get:
+      summary: |
+        Get list of links to annual cycle timeseries for standard climatology
+        time periods.
+      description: |
+        Get list of links to annual cycle timeseries for standard climatology periods for specified streamflow. Standard climatology periods are:
+
+        - 1961-1999
+        - 1971-2000
+        - 1981-2010
+        - 2010-2039
+        - 2040-2069
+        - 2070-2099
+
+        The list of links returned are for those periods, limited by the time period spanned by the streamflow result set.
+      tags:
+        - Results
+      parameters:
+        - $ref: '#/components/parameters/resultId'
+      responses:
+        200:
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/StreamflowResultAnnualCycleStandardPeriodList'
+        404:
+          $ref: '#/components/responses/404NotFound'
+
   /streamflow/results/{id}/annual-max:
     get:
       summary: Get annual maximum timeseries for specified streamflow
@@ -805,8 +835,10 @@ components:
             uri: 'https://services.pacificclimate.org/streamflow/results/6'
           - rel: 'deprecate'
             uri: 'https://services.pacificclimate.org/streamflow/results/6'
-          - rel: 'annual-cycle'
+          - rel: 'annual-cycle-template'
             uri: 'https://services.pacificclimate.org/streamflow/results/6/annual-cycle/{startYear}-{endYear}'
+          - rel: 'annual-cycle-standard-periods'
+            uri: 'https://services.pacificclimate.org/streamflow/results/6/annual-cycle/standard-periods'
           - rel: 'annual-max'
             uri: 'https://services.pacificclimate.org/streamflow/results/6/annual-max'
           - rel: 'annual-mean'
@@ -850,6 +882,52 @@ components:
           start_date: '2000-01-01'
           end_date: '2100-12-31'
 
+    # List of streamflow result annual cycles for standard climatology periods
+
+    StreamflowResultAnnualCycleStandardPeriodsListLitem:
+      description: Hyperlink to annual cycle data for specified period.
+      type: object
+      allOf:
+        - type: object
+          properties:
+            startYear:
+              type: integer
+            endYear:
+              type: integer
+          required:
+            - startYear
+            - endYear
+        - $ref: '#/components/schemas/ListItemUri'
+      example:
+        startYear: 1971
+        endYear: 2000
+        uri: 'https://services.pacificclimate.org/streamflow/results/6/annual-cycle/1971-2000'
+
+    StreamflowResultAnnualCycleStandardPeriodList:
+      description: |
+        List of hyperlinks to annual cycle data for standard climatology periods.
+      type: array
+      items:
+        $ref: '#/components/schemas/StreamflowResultAnnualCycleStandardPeriodsListLitem'
+      example:
+        - startYear: 1961
+          endYear: 1999
+          uri: 'https://services.pacificclimate.org/streamflow/results/6/annual-cycle/1961-1999'
+        - startYear: 1971
+          endYear: 2000
+          uri: 'https://services.pacificclimate.org/streamflow/results/6/annual-cycle/1971-2000'
+        - startYear: 1981
+          endYear: 2010
+          uri: 'https://services.pacificclimate.org/streamflow/results/6/annual-cycle/1981-2010'
+        - startYear: 2010
+          endYear: 2039
+          uri: 'https://services.pacificclimate.org/streamflow/results/6/annual-cycle/2010-2039'
+        - startYear: 2040
+          endYear: 2069
+          uri: 'https://services.pacificclimate.org/streamflow/results/6/annual-cycle/2040-2069'
+        - startYear: 2070
+          endYear: 2099
+          uri: 'https://services.pacificclimate.org/streamflow/results/6/annual-cycle/2070-2099'
 
     # Streamflow result data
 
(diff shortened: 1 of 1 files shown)


streamflow-api.yaml
```diff
@@ -0,0 +1,768 @@
...
+
+    ### Streamflow orders
+
+    # Individual streamflow orders
+
+    StreamflowOrderStateNew:
+      description: Defines a new order for a streamflow result set.
+      type: object
+      properties:
+        hydromodel_output_id:
```
rod-glover: I _think_, but am not sure, that the way to indicate the hydromodel output desired as input to this computation is with such an id. This has the virtue of simplicity but it exposes an internal implementation detail, and that is regarded as undesirable in resource-oriented architectures (RESTful architecture).

The alternative would be the hyperlink to the hydromodel output. This does not expose an internal, but it does require the backend to translate the hyperlink to an actual hydromodel output record, which is more roundout.

I'm not sure which to choose.

streamflow-api.yaml
```diff
@@ -0,0 +1,768 @@
...
+          example: -123.5
+        latitude:
+          type: number
+          example: 48.5
+        notification:
+          type: object
+          properties:
+            method:
+              type: string
+              enum: [email]
```
rod-glover: Add `none` option?

streamflow-api.yaml
```diff
@@ -0,0 +1,768 @@
...
+      properties:
+        hydromodel_output_id:
+          $ref: '#/components/schemas/HydromodelOutputId'
+        longitude:
+          type: number
+          example: -123.5
+        latitude:
+          type: number
+          example: 48.5
+        notification:
```
rod-glover: Should make this `notifications: [ ... list ... ]` ? Is there a plausible use case for multiple notifications?

rod-glover:
An overarching question: Should the list resources (e.g., `/streamflow/orders`) return less metadata than the individual item resources (e.g., `/streamflow/orders/4`). Alternately stated: Should the individual item resources contain more info?

I'd like to review all such resources with an eye to this question and the general question of whether the representations for each type of resource is yet complete.

corviday:
My approach has been to assume that the resource collection endpoints provide all information needed to answer the question: **which resource am I looking for?** And that the resource documents return all information needed to answer the question: **I've now selected this resource, what information is available about it?**

Maybe this is the backwards approach, though, since what information is needed to select the resource of interest varies based on how one expects the information to be used or what one expects the front end to be like.

One could, for example, imagine multiple UIs allowing a user to select a hydromodel output dataset:

1. Possible UI 1: the user selects which GCM and emissions scenario they are interested in from dropdown menus, the spatial coverage of that hydromodel output dataset is displayed on a map, then the user picks a point they are interested in from a map. For this scenario, you'd want the hydromodel output resource collection to return GCM and emissions scenario information, but since you don't display the hydromodel output on a map until after the user picks one, you could have grid info on the resource document, and not in the resource collection.

2. Possible UI 2: the user picks a point on a map, and then gets a list of all the available GCM/scenario combinations  of hydromodel output that could provide data at that map point. In this case, you'd want the hydromodel output collections resource to contain spatial coverage information, but individual resource documents would have the GCM and scenario.

(I've been assuming our UI will look more like the former, because I think it would be easier to design a UI like that, but the latter might be more in line with how the engineers want to approach things; they probably start out knowing exactly what point in space they want to get information about, and don't care as much about available models.)

streamflow-api.yaml
```diff
@@ -0,0 +1,768 @@
...
+              $ref: '#/components/schemas/HydromodelOutputId'
+            model:
+              type: string
+              example: CanESM2
+            experiment:
+              type: string
+              example: rcp85
+            etc:
+              type: string
+              example: ... more tbd ...
```
corviday: Spatial coverage info is needed, I think, so that it's possible to tell whether a result for a given point can be generated from this hydromodel output. 

Something like minimum and maximum latitude and longitude, as well as latitude and longitude cell size would be enough to provide context to the cell_x and cell_y values in the result endpoint.

corviday: Yuck, it looks like hydromodel output files have irregular spatial extents:
spatialcoverage

So in order to provide enough information to determine whether a result file for  given point can be made from this hydromodel output, we may need to return a geoJSON polygon or something as the spatial output., not just a bounding box. That's a bit of a headache.

streamflow-api.yaml
```diff
@@ -0,0 +1,768 @@
...
+          latitude: 50.2
+          notification:
+            method: email
+            address: def@example.ca
+
+    ### Streamflow results
+
+    # Individual streamflow results
+
+    StreamflowResultState:
```
corviday: Temporal coverage information would be useful. A start date, a stop date, and a resolution, maybe? Temporal coverage info would mean you could actually figure out what dates to request for the annual cycle resource.

streamflow-api.yaml
corviday: Hm, maybe if we need to provide actual spatial coverage, it could be a separate resource, like maybe, like `/streamflow/hydromodel_output/4/bounds`? We probably don't want to calculate it until a client is specifically looking for it, though caching could be a big help, it's not like the coverage of a netCDF file is going to change.

streamflow-api.yaml
```diff
@@ -0,0 +1,768 @@
...
+                - rel: 'self'
+                  uri: 'http://apis.paccifcclimate.org/streamflow/orders/5'
+                - rel: 'cancel'
+                  uri: 'http://apis.paccifcclimate.org/streamflow/orders/5'
+                - rel: 'hydromodel_output'
+                  uri: 'http://apis.paccifcclimate.org/hydromodel_outputs/4'
+                - rel: 'results/metadata'
+                  uri: 'http://apis.paccifcclimate.org/streamflow/results/6'
+                - rel: 'results/annual-max'
+                  uri: 'http://apis.paccifcclimate.org/streamflow/results/6/annual-max'
```
corviday: So it's the RESTy thing to do to include these links to annual cycle, annual mean, and annual max even though they don't exist when the order is created and will be 404s?

streamflow-api.yaml
```diff
@@ -0,0 +1,768 @@
...
+        links:
+          allOf:
+            - $ref: '#/components/schemas/HypermediaLinkList'
+            - example:
+                - rel: 'self'
+                  uri: 'http://apis.paccifcclimate.org/streamflow/orders/5'
+                - rel: 'cancel'
+                  uri: 'http://apis.paccifcclimate.org/streamflow/orders/5'
+                - rel: 'hydromodel_output'
+                  uri: 'http://apis.paccifcclimate.org/hydromodel_outputs/4'
```
corviday: URI should probably be http://apis.pacificclimate.org/streamflow/hydromodel_outputs/4

rod-glover:
@corviday, I think your reasoning (first 2 paragraphs) is right on. It matches what I understand of resource-oriented/RESTful design, and my own opinions.

I think the "support multiple UIs" principle is also important. You want UIs to be able to slice and dice information in various ways (like your example). It's an anti-pattern to have endpoints that are tuned for specific views of the information; instead you want endpoints that deliver broader information that the UI can present in various ways.

This does argue for collection endpoints that deliver as much metadata as the individual endpoints. Given that our endpoints are pretty much 100% metadata, aside from the explicitly data endpoints, I think that clinches it.

streamflow-api.yaml
rod-glover: I agree completely. Caching will make a very big difference here, but it still may be that it's better to provide the bounds at a separate endpoint. At this stage, I prefer that. 

(BTW, the bounds would likely be a kind of information not supplied in the list resource that would be present in the hypothetical item resources. But partitioning it as a separate endpoint eliminates that question.)

rod-glover: Will we want to add filtering (to the list resource) by location? E.g., list all hydromodels that contain this location? Or some spatial extent?

streamflow-api.yaml
rod-glover: Would that temporal coverage also (or alternatively) be part of the hydromodel_output, which I believe determines the temporal coverage of the computed result?

(There's no reason we can't add temporal metadata to both. I tend to want to not duplicate information, but this isn't a database that should be normalized, but information resources that are meant to be useful!)

rod-glover: Also, what do you recommend as the resolution datum? 
- A string enum like daily, monthly, yearly?
- An exact interval specified either in seconds or by an ISO8601 duration string?

streamflow-api.yaml
rod-glover: Great observation. REST has nothing to say about this either way. A link to an absent resource isn't really non-RESTful, but it's not really very logical either. A good rule is don't have links that you can know in advance are invalid.

streamflow-api.yaml
corviday: I think putting temporal coverage in either the result metadata or the hydromodel output metadata would be fine, as long as it's accessible somewhere. It is theoretically possible to generate a result with a shorter timespan than the hydromodel output, but I can't imagine why we'd want to, and don't see any reason to build in support for that possibility out of the gate.

------

Upon further reflection, I think I was wrong. The result metadata does not need a time resolution attribute because I can't see any reason a client would need that information. A client never works with the raw result data and doesn't need details of its representation in the datafile; the client only interacts with the data in the form of the aggregated annual-mean, annual-max, and annual-cycle endpoints.

streamflow-api.yaml
corviday: Hm. Do you think it's better to do that sort of filtering in the back end or the front end?

streamflow-api.yaml
rod-glover: I intentionally named the hydromodel resources as `/hydromodel_outputs` and not `/streamflow/hydromodel_outputs` because they aren't necessarily associated only with streamflow. Streamflow uses them, but so could other things.

Alternative viewpoint is that these resources are tuned to streamflow's needs, and we'd not want other concerns changing the API and breaking streamflow app(s). 

What do you think?

corviday: I guess either way seems justifiable to me. No strong preference.

streamflow-api.yaml
```diff
@@ -0,0 +1,891 @@
...
+        401:
+          $ref: '#/components/responses/401Unauthorized'
+        404:
+          $ref: '#/components/responses/404NotFound'
+        409:
+          $ref: '#/components/responses/409Conflict'
+
+
+  /streamflow/results/{id}/annual-cycle/{startYear}-{endYear}:
+    get:
```
corviday: A different way to handle the annual cycle query that just occurred to me, if one of the REST design goals is to not make the client construct arbitrary URLs: 

- `/streamflow/result/0 `includes a link to `/streamflow/result/0/annual-cycle` (along with other metadata)
- `/streamflow/result/0/annual-cycle` is solely a collection of links to annual cycles for valid climatologies present in this data, so it would have links to, eg:
    - `/streamflow/result/0/annual-cycle/1961-1999`
    - `/streamflow/result/0/annual-cycle/1971-2000`
    - `/streamflow/result/0/annual-cycle/1981-2010`
    - `/streamflow/result/0/annual-cycle/2010-2039`
    - `/streamflow/result/0/annual-cycle/2040-2069`
    - `/streamflow/result/0/annual-cycle/2070-2099`
- Pages like` /streamflow/result/0/annual-cycle/2070-2099` would actually return a timeseries.

Or there could be a climatology resource type, like` /streamflow/result/0/climatology/1960-1989/annual-cycle` or something, but I don't know of any other data analysis we'd like broken out by climatology, so I don't think that buys us anything.

rod-glover: You rock. This is an excellent idea.

rod-glover: This presumes that one knows in advance what climatologies are of interest. But it can't hurt to have a standard set listed.

streamflow-api.yaml
corviday: I'm having a hard time thinking of a circumstance where we'd want the none option.

rod-glover: Yeah, and although it is easily circumvented, it is a way to attribute orders to users.

corviday:
I did a final readthrough of the API just now, and I think it looks great! 

The only thing I noticed is that I think we've made the `cell_x_index` and `cell_y_index` attributes of `result` superfluous: a client doesn't need to care about cell indices in the netCDF file - a client cares about:
1. whether the point it is interested in corresponds to an existing result, which it can determine with `cell_x`, `cell_y`, `cell_x_size`, and `cell_y_size` on `result`
2.  if not, whether the point it is interested in inside a the spatial coverage of a particular `hydromodel_output`, which it can determine with `hydromodel_output/{id}/bounds`

There's no longer much use for `cell_x_index` or `cell_y_index`, though it doesn't do any harm to have them.

(I can, however, imagine scenarios in which a client might still want to know `cell_x_size` and `cell_y_size` on `hydromodel_output`; I can imagine a client that cares about the _resolution_ of a dataset even if it doesn't have to calculate an exact cell itself.)

rod-glover:
Good point. I'll remove the cell indices; leaner is better.

rod-glover:
I think this is it for the API design ... except for security (user authentication and authorization). But that's a big enough issue that I'll create a separate issue and PR (if needed) for it.

rod-glover:
Ack. API versioning. Which is [a fraught subject](https://www.troyhunt.com/your-api-versioning-is-wrong-which-is/). But important to consider and possibly do regardless.

rod-glover:
Re. API versioning. Definitely worth reading the article I mentioned above, [Your API versioning is wrong](https://www.troyhunt.com/your-api-versioning-is-wrong-which-is/). 

And it's even more worth reading the often very well-considered responses to that article. The general sentiment there seems to be that putting versioning in the resource URIs (e.g., `/api/v2/...`), however "wrong" it may be, is the most pragmatic approach and arguably not all that wrong anyway. The article's author [agrees](https://www.troyhunt.com/your-api-versioning-is-wrong-which-is/#comment-2637822944). I agree too.

So assuming we _all_ agree on that, the question is whether to introduce versioning (in the URIs) now, or wait until there is actually another version. If we do that, then URIs sans version should default to v1 (to preserve compatibility) and anyone who wants v2 or later must specify so in the URIs.

corviday:
That all seems reasonable to me.

rod-glover:
The failing tests have nothing to do with the API spec. Gonna merge.

PR #92, opened 2018-09-12
Update API to interface with 

corviday:
Minor modifications so that the existing backend queries will run against a database updated to store streamflow data. Pretty much just changing all instances of `DataFileVariable` to `DataFileVariableGridded`.

Posted for review and testing, but we may not want to merge it until we're ready to update the live database.

Test instance is [here](http://docker-dev02.pcic.uvic.ca:30066/#/).

### DIFF
--- a/ce/api/data.py
+++ b/ce/api/data.py
@@ -4,7 +4,7 @@
 import numpy as np
 
 from modelmeta import Run, Emission, Model, TimeSet, DataFile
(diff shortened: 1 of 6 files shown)


corviday:
This PR updates the backend APIs to interface with the database once the streamflow-related tables are added, but at this point, I think redoing the small number of changes from scratch would be easier than trying to merge this.

PR #95, opened 2019-02-07
Fix API Documentation

### LINKED ISSUE #94: Make using the API easier
There are a few minor things about the API code that can trip up a new developer writing code than runs against the API and could be easily fixed:

- [x] The default ensemble on queries that require one is `ce`, which no longer exists. Should be `ce_files`
 - [x] Documentation on many queries still describes the time argument as a value from 1 to 17, when queries have been updated to accept a resolution and an index
 - [x] Overall documentation review, make sure it's all up to date

nikola-rados:
Resolves #94 

The docs were slightly out of date for the ```time``` and ```timescale``` args in ```multistats.py``` and ```stats.py```.  Furthermore, the default ensemble name has been updated to ```ce_files```.  Other than those small things the docs looks great!

### DIFF
--- a/ce/api/multistats.py
+++ b/ce/api/multistats.py
@@ -4,7 +4,7 @@
 from ce.api.stats import stats
 from ce.api.util import search_for_unique_ids
 
-def multistats(sesh, ensemble_name='ce', model='', emission='', time=0,
+def multistats(sesh, ensemble_name='ce_files', model='', emission='', time=0,
                area=None, variable='', timescale=''):
     '''Request and calculate statistics for multiple models or scenarios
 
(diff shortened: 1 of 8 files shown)


nikola-rados:
@corviday if you wouldn't mind giving this a once over just to make sure I didn't miss anything!

ce/api/multistats.py
```diff
@@ -4,7 +4,7 @@
 from ce.api.stats import stats
 from ce.api.util import search_for_unique_ids
 
-def multistats(sesh, ensemble_name='ce', model='', emission='', time=0,
+def multistats(sesh, ensemble_name='ce_files', model='', emission='', time=0,
```
corviday: Is 0 an invalid default argument? Is this one of the queries that will return an error if it actually receives a 0 here?

nikola-rados: Ah right.  So that issue occurs in `stats.py`:
```
    ...
    if time: #HERE
        try:
            time = int(time)
        except ValueError:
            raise Exception('time parameter "{}" not convertable to an integer.'
                            .format(time))
    else:
        time = None
    ...
    query = sesh.query(Time.timestep).filter(Time.time_set_id == df.timeset.id)
    if time: #HERE
        query.filter(Time.time_idx == time)
    timevals = [ t for t, in query.all() ]
    timeval = mean_datetime(timevals)
    ...
```
I think changing `if time` to `if time is not None` should take care of the problem?  Since we are already converting it to an int, it shouldn't matter is we pass in `'0'` or `0`.

corviday: Sounds good. Can you add that change to this PR?

corviday: After further discussion, we realized that making this change would be complicated enough to warrant a separate PR.

PR #96, opened 2019-03-15
Feature cell method filter

nikola-rados:
In order to distinguish between **standard deviation datasets** and **mean datasets**, some extra filtering had to be done in the api.  Based on the `cell_method` parameter, a regex pattern is selected and used to match with the modelmeta `DataFileVariable.variable_cell_methods` attribute.  All of the metadata containing `standard_deviation` adheres to the [CF Metadata Conventions](http://cfconventions.org/cf-conventions/cf-conventions.html#cell-methods).  However older climatological mean files did not always update `cell_methods`, therefore they contain a wider variety of possible values.  I've elected to use separate regex patterns because of this.

Some test cases were altered/added to cover the new `cell_method` filter.

Pinned the version of sqlalchemy to `1.2.18`, as the `1.3` release was causing issues with multiple joins (perhaps this is something we should look at?).


ce/api/multistats.py
```diff
@@ -5,7 +5,7 @@
 from ce.api.util import search_for_unique_ids
 
 def multistats(sesh, ensemble_name='ce_files', model='', emission='', time=0,
-               area=None, variable='', timescale=''):
+               area=None, variable='', timescale='', cell_method=''):
```
jameshiebert: Maybe we should include `mean` as the default value for `cell_method`? That would be backward compatible, right?

ce/api/util.py
```diff
@@ -114,4 +122,18 @@ def search_for_unique_ids(sesh, ensemble_name='ce', model='', emission='',
     if timescale:
         query = query.filter(mm.TimeSet.time_resolution == timescale)
 
+    if cell_method:
```
jameshiebert: the `else` branch of this, would I guess be that it doesn't filter on `cell_method`. But I think that it would be better to be explicit. Always filter on `cell_method`, but the default is `mean`. If, historically, we haven't populated the `cell_method` attribute, what if we simply populated everything with `mean` right now?

ce/api/util.py
```diff
@@ -114,4 +122,18 @@ def search_for_unique_ids(sesh, ensemble_name='ce', model='', emission='',
...
+
+        cell_methods = sesh.query(mm.DataFileVariable.variable_cell_methods)\
+                        .distinct(mm.DataFileVariable.variable_cell_methods).all()
+        pattern = {
+            'standard_deviation': r'time: [a-z]* time: standard_deviation over days',
+            'mean': r'time: [a-z]*( time: mean .*)?'
+        }[cell_method]
+
+        matching_cell_methods = [r[0] for r in cell_methods if re.match(pattern, r[0])]
+        query = query.filter(mm.DataFileVariable.variable_cell_methods.in_(matching_cell_methods))
```
jameshiebert: FWIW, you can use regular expressions directly in PostgreSQL with the `~` operator. In SQLAlchemy, this looks something like this:

```
pattern = ...
session.query(Object).filter(Object.column.op('regexp')(pattern))
```

This would cut your queries in half and make the code more clear and direct.

requirements.txt
```diff
@@ -7,6 +7,7 @@ numpy
 netcdf4==1.3
 GDAL~=2.3
 rasterio~=1.0
+sqlalchemy==1.2.18
```
jameshiebert: Yes, I would definitely support pinning the SA version in general, despite what every problems you are having with v1.3+

ce/tests/conftest.py
```diff
@@ -231,14 +231,19 @@ def make_data_file_variable(file, var_name=None, grid=grid_anuspline):
         var_name_to_alias = {
             'tasmin': tasmin,
             'tasmax': tasmax,
-        }
+        }[var_name]
+        variable_cell_methods = {
```
jameshiebert: Is it worth adding a test here that has a cell_method "mean over days" to ensure that the code works into the future when we properly attribute our normal cases?

ce/api/util.py
```diff
@@ -95,8 +96,15 @@ def mean_datetime(datetimes):
     mean = np.mean(timestamps)
     return datetime.fromtimestamp(mean, tz=timezone.utc)
 
+def validate_cell_method(cell_method):
```
jameshiebert: This is very explicit, which in this case I like :) It would save a couple lines until we support methods, but you could probably also write the function body as:
```
return cell_method in ('mean', 'standard_deviation')
```
and it would be equally explicit, but also short.

ce/api/util.py
nikola-rados: The `cell_method` attribute is always populated.  The issue is that it hasn't been populated with the operation done to produce the climatologies.  

For example, an old `tasmax` climo file would say `time: maximum`, which describes `tasmax` itself but leaves out any further detail about what has been done with that data.  In order to describe the statistical operation it would have to say `time: maximum time: mean over days`, which is what we now do.  If we were to simply append `time: mean over days` to all the old climo files I think that would be fine.  

I agree with you though, that it would be a good idea to use `mean` as a default, since leaving it blank would give us a mix of `standard_deviation` and `mean` data.

ce/api/util.py
nikola-rados: I had tried that (and a bunch of alternatives) but I could never get it to work.  From what I had found it wasn't supported, hence why I created a separate query.  If I'm missing something I'd be happy to filter that way instead but I haven't been able to get it to work.

ce/tests/conftest.py
nikola-rados: Yes, that would be a good idea!

ce/api/util.py
```diff
@@ -95,14 +96,30 @@ def mean_datetime(datetimes):
...
-                          variable='', time=0, timescale=''):
+                          variable='', time=0, timescale='', cell_method='mean'):
+    if not validate_cell_method(cell_method):
+        raise Exception('Unsupported cell_method: {}'.format(cell_method))
+
+    cell_methods = sesh.query(mm.DataFileVariable.variable_cell_methods)\
+                    .distinct(mm.DataFileVariable.variable_cell_methods).all()
+    pattern = {
+        'standard_deviation': r'time: [a-z]* time: standard_deviation over days',
+        'mean': r'time: [a-z]*( time: mean over days)?'
```
rod-glover: These regexes aren't robust to whitespace. I don't think there is anything in the CF standard that says the space between words must be exactly one space. Suggest you make it more robust by allowing `\s+` between words.

rod-glover: Testing robustness would be nice but is probably not necessary.

ce/api/util.py
rod-glover: Yes, we need to preserve the results that existing web API clients (i.e., CE) get with their existing hits on those endpoints. If not, we have to modify all clients. Better to avoid that if possible.

nikola-rados:
@rod-glover are you happy with where it is at now?

rod-glover:
LGTM

ce/api/util.py
```diff
@@ -95,14 +96,30 @@ def mean_datetime(datetimes):
...
-                          variable='', time=0, timescale=''):
+                          variable='', time=0, timescale='', cell_method='mean'):
+    if not validate_cell_method(cell_method):
+        raise Exception('Unsupported cell_method: {}'.format(cell_method))
+
+    cell_methods = sesh.query(mm.DataFileVariable.variable_cell_methods)\
+                    .distinct(mm.DataFileVariable.variable_cell_methods).all()
+    pattern = {
+        'standard_deviation': r'time:[a-z\s]*time:\s+standard_deviation\s+over\s+days',
+        'mean': r'time:[a-z\s]*\s+(time:\s+mean\s+over\s+days)?'
```
corviday: As just discussed in meeting with @nikola-rados, @rod-glover, and @corviday , `generate_climos` [is going to start putting out some climatologies with cell_methods that end "mean over years"](https://github.com/pacificclimate/climate-explorer-data-prep/issues/64) and this regex should support "years" in addition to "days" here.

nikola-rados: I'll add it in!

nikola-rados:
@corviday are you happy with the changes?

Issue #98, opened 2019-04-11
cell_method filter match() vs fullmatch()


nikola-rados:
I've discovered that the `cell_method` filter is not working as intended.  It uses `re.match()` to check for desired `cell_methods` but will return `True` for those with only a partial match.  Change it to `re.fullmatch()`to ensure we aren't getting mixed data. 

This will need to be tagged and released, likely `1.1.1`.

nikola-rados:
Fixed in #99

PR #99, opened 2019-04-12
Fix cell_methods filter

nikola-rados:
The `cell_methods` filter I implemented in PR #96 did not function as I intended.  It would grab `standard_deviation` data file ids while filtering for `mean` data file ids which was not a desirable behaviour.  The solution was to add some more structure to the logic of the filter.  Any climatological standard deviations will have an expected `cell_method` that we can regex for rather simply.  However the climatological means do not consistently use `cell_method` correctly.  To fix this I added a pattern to match for the correct `cell_method` and ignore other cases.  These ignored cases would be checked against the filter for `standard_deviation` and if they did not match they would be considered means.  This resolves issue #98.

### DIFF
--- a/ce/api/util.py
+++ b/ce/api/util.py
@@ -99,6 +99,20 @@ def mean_datetime(datetimes):
 def validate_cell_method(cell_method):
     return cell_method in ('mean', 'standard_deviation')
 
+def find_matching_cell_methods(cell_methods, target_method):
+    def filter_on_method(cell_method, target_method):
+        pattern = r'time:[a-z\s]*time:\s+{}\s+over\s+(days|years)'.format(target_method)
+        return re.fullmatch(pattern, cell_method)
+
+    if target_method == 'mean':
+        return [cell_method for cell_method in cell_methods
+                if filter_on_method(cell_method, target_method) or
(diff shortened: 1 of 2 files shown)


ce/api/util.py
```diff
@@ -99,19 +99,29 @@ def mean_datetime(datetimes):
 def validate_cell_method(cell_method):
     return cell_method in ('mean', 'standard_deviation')
 
+def find_matching_cell_methods(cell_methods, target_method):
+    def filter_on_method(cell_method, target_method):
+        pattern = r'time:[a-z\s]*time:\s+{}\s+over\s+(days|years)'.format(target_method)
+        return re.fullmatch(pattern, cell_method)
+
+    if target_method == 'mean':
```
corviday: This bit desperately needs an explanatory comment as to _why_ mean has this weird special case handling. Since it's for weird historical reasons, it's definitely _not_ something someone could figure out by reading the code and using logic. Explain what it does and why it has to be done to the future programmer who is looking at this code for the first time.

nikola-rados: Ok good idea, I added a comment

nikola-rados:
Let me know if you think anything else should be changed @corviday, otherwise I think this is ready.

PR #102, opened 2019-05-22
Modifies to geo cache to maintain its own size stats

### LINKED ISSUE #101: Optimize cache code
In preparation for using the `climate-explorer-backend` to drive [Plan2Adapt](https://plan2adapt.ca) (P2A), I've been taking a look at the performance of the various queries.

P2A makes hundreds (slightly over a thousand) calls to the `/stats` API calls, so small slow downs add up to a lot.

Turns out that the speed of the cache is slow. Like *really* slow. Like to the point where it would be faster if we simply turned *off* the cache.

Why? In order to control the size of the cache, it [recursively counts the size](https://github.com/pacificclimate/climate-explorer-backend/blob/master/ce/api/geo.py#L26) of every object it contains. Doing this over and over adds up to a huge performance penalty. Over the course of 1 P2A run, simply measuring the size of the cache took around 43% of the run time while fetching the actual data from disk only took 21% of the run time.

I think that it would be fairly easy to have the cache keep track of its own size and just increment/decrement  when something is added/expired. That would presumably give us an almost 2x speedup.

jameshiebert:
As documented in #101 the geo cache spends an inordinate amount of time simply recursively counting the size of its own objects (on every cache miss). This takes up a large percentage of the run time when seeding the cache with new data.

The patch fixes the problem by only deeply sizing objects on cache insertion and expiry. Performance testing indicates that this has reduced the runtime of this section of code from 16 minutes (and a P2A run) to < 1 second. Win!

Resolves #101

### DIFF
--- a/ce/api/geo.py
+++ b/ce/api/geo.py
@@ -3,6 +3,7 @@ import sys
 import math
 from collections import OrderedDict
 from threading import RLock
+from contextlib import ContextDecorator
 
 from netCDF4 import Dataset
 import numpy as np
@@ -52,6 +53,22 @@ def getsize(obj):
 
 log = logging.getLogger(__name__)
 cache_lock = RLock()
+
+
+class cache_sizer(ContextDecorator):
+    def __init__(self, cache):
+        self.cache = cache
+
+    def __enter__(self):
+        self.startsize = sys.getsizeof(self.cache)
+        return self
+
+    def __exit__(self, *exc):
+        self.endsize = sys.getsizeof(self.cache)
+
+    def delta_size(self):
+        return self.endsize - self.startsize
+    
 
 class memoize(object):
     '''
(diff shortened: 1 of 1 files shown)


ce/api/geo.py
```diff
@@ -106,13 +123,18 @@ def __call__(self, *args):
                     result = self.func(*args)
 
                 with cache_lock:
-                    self.cache[key] = result
+                    with cache_sizer(self.cache) as sizer:
+                        self.cache[key] = result
+                    self.size += getsize(result) + sizer.delta_size()
```
rod-glover: OK, I learned something here. Python has a non-intuitive (to me) semantics for the scope of a variable defined in a `with` statement. It persists. Tricky. I guess my comment here is: this code is tricky. But I get what you are doing.

ce/api/geo.py
```diff
@@ -106,13 +123,18 @@ def __call__(self, *args):
                     result = self.func(*args)
 
                 with cache_lock:
```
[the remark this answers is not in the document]
jameshiebert: Hrm, not sure. One for a hit and one for a miss, I assume, but yes, functionally it would probably be the same with a single `with` block.

ce/api/geo.py
jameshiebert: Yeah, that definitely trips people up. There is no `with` scope.

Context managers are there to do something before and/or after a block of code. The obvious use case is a file open (where you want the file to always be closed when the block ends). But after the block ends, even if the file is closed, the file object still has value to the rest of the code. You can check its `filename` property, you list the file descriptor, test whether it's open or not, etc. But to do those things, the object has to still exist.

PR #114, opened 2019-12-17
Add sphinx documentation of API functionality

corviday:
Adds sphinx for API documentation. 

Because we prefer Markdown to reStructured Text, almost all text is written in markdown files that are then included into the rST files with `m2r`, and sphinx's native rST files are used for structure only. The exceptions are the main index file, and the help for the `metadata` endpoint, which takes advantage of functionality only available in rST to put a big red warning box on the page.

There are no real changes to any files outside the newly created folder `/doc`, just whitespace formatting on docstrings of API endpoint functions to make them display better in sphinx, and transferring a docstring to a different function in `watershed.py`.


doc/source/overview.md
```diff
@@ -0,0 +1,32 @@
+# Overview
+
+## Raster Data
```
rod-glover: The term "raster" I believe refers to an image that is rendered by scanning as in a CRT. Maybe it's just me, but I dislike this term. "Grid" or "gridded" seems preferable to me. Can we change that?

doc/source/overview.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Data Access
+This system is not designed to provide individual access to every point in the raster data matrix, but calculates and makes available summary data over user-defined spatial and temporal scales to support analysis and visualization. All numerical API endpoints accept a spatial area of interest (specified in URL-encoded WKT as a point or polygon) and return the mean of all cells within the designated area. If no area is designated, the mean across the entire spatial extent of the dataset is returned. 
+
+Datafiles are organized by temporal range and resolution.
+
+## Modeled Data
+
+Most data accessible by the API is climate data from a General Circulation Model (GCM) or Regional Climate Model (RCM). These numerical models represent physical processes in the atmosphere, ocean, cryosphere, and land surface of the earth. They simulate the response of the global climate system. The models output meteorological projections like temperature and precipitation. Other available variables represent statistical indices derived from the meteorological outputs.
+
```
rod-glover: > The models output meteorological projections like temperature and precipitation.

How about

Model outputs are meteorological variables such as temperature and precipitation.

doc/source/overview.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Data Access
+This system is not designed to provide individual access to every point in the raster data matrix, but calculates and makes available summary data over user-defined spatial and temporal scales to support analysis and visualization. All numerical API endpoints accept a spatial area of interest (specified in URL-encoded WKT as a point or polygon) and return the mean of all cells within the designated area. If no area is designated, the mean across the entire spatial extent of the dataset is returned. 
+
+Datafiles are organized by temporal range and resolution.
+
+## Modeled Data
+
+Most data accessible by the API is climate data from a General Circulation Model (GCM) or Regional Climate Model (RCM). These numerical models represent physical processes in the atmosphere, ocean, cryosphere, and land surface of the earth. They simulate the response of the global climate system. The models output meteorological projections like temperature and precipitation. Other available variables represent statistical indices derived from the meteorological outputs.
```
rod-glover: > The models output meteorological projections like temperature and precipitation.

How about

Model outputs are meteorological variables such as temperature and precipitation.

doc/source/overview.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Emissions Scenarios
+
+Each dataset has an associated emissions scenario, sometimes called `experiment`. Emissions scenarios represent a range of possible future projections for greenhouse gas concentrations. GCMs and RCMs are typically driven by historical data on greenhouse gas concentrations when modeling the past and present, before using an emissions scenario to provide input on greenhouse gas concentrations to drive future projections.
```
rod-glover: > sometimes called `experiment`

I think there are two things in one here:

- The scenario is sometimes called an "experiment," in the sense that a model run with a given scenario represents an experiment with the model earth system.
- We use the parameter name `experiment` to designate an emissions scenario in code/API.

doc/source/overview.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Runs
+
+Each dataset has an associated run string, sometimes called `ensemble_member`, which represents the initialization settings of the simulating model. A GCM or RCM may be run multiple times with different initialization conditions.
```
rod-glover: Similarly to "experiment," let's define terms here: A set of related runs with different initializations of the same model is called an "ensemble" (in reference to the notion of statistical ensemble).

Should we also mention (and link to more info on) the R-I-P code for this?

doc/source/overview.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Emissions Scenarios
+
+Each dataset has an associated emissions scenario, sometimes called `experiment`. Emissions scenarios represent a range of possible future projections for greenhouse gas concentrations. GCMs and RCMs are typically driven by historical data on greenhouse gas concentrations when modeling the past and present, before using an emissions scenario to provide input on greenhouse gas concentrations to drive future projections.
+
+### Runs
+
+Each dataset has an associated run string, sometimes called `ensemble_member`, which represents the initialization settings of the simulating model. A GCM or RCM may be run multiple times with different initialization conditions.  
+
+## Climatological Means
```
rod-glover: Do we need to talk more generally about climatological statistics, e.g., climo mean, std deviation, median, quantiles, etc.? We have a few of these, yes?

doc/source/overview.md
corviday: I was trying to get at the netCDF data model of a cube or hypercube without getting into the details of the netCDF format, which, despite being not directly accessible to the API user, do affect how data is organized. I don't like "gridded" because it implies two-dimensionality. Maybe title this section, uh, Data Matrix?

doc/source/workflow.md
```diff
@@ -0,0 +1,29 @@
+# Typical Workflows
+
+## List Available Datasets
+
+Datafiles are organized into ensembles, which collect all data needed for a specific purpose. To view a list of all datasets in a particular ensemble, query the `multimeta` API. The `multimeta` API gives each datafile as a unique identifier string and a collection of attributes describing the data contained within that file. After a datafile of interest has been determined from its metadata attributes, its unique identifier string may be used to request the data.
```
rod-glover: Let's acknowledge:

(The term "ensemble" is a misnomer in this usage; a more appropriate term would be "collection." For historical reasons the term "ensemble" is embedded in the code and it is not easily changed at this point.)

rod-glover: Hyperlink to `multimeta` documentation?

doc/source/workflow.md
```diff
@@ -0,0 +1,29 @@
...
+
+Datasets with `multi_year_mean` equal to `false` represent nominal time datasets; no aggregation has been done between years. A monthly dataset covering 1961-1990 would feature 360 timestamps.
+
+## Request Numerical Data From A Climatological Mean Datafile
+
+The `timeseries` endpoint returns a timeseries with the mean value of every timestamp in the datafile. It requires a datafile's unique identification string, and optionally a spatial area of interest defined as a Well Known Text (WKT) Polygon or Point. For a climatological mean datafile, the resulting timeseries represents an average annual cycle over the period described by the dataset. The annual cycle may have twelve monthly values, four seasonal values, or a single annual value.
```
rod-glover: Hyperlink to `timeseries` docu?

doc/source/workflow.md
```diff
@@ -0,0 +1,29 @@
...
+
+The `stats` endpoint returns statistical measures (`mean`, `stdev`, `min`, `max`, and `median`) of a single dataset identified by its unique identification string, a temporal index, and optionally a spatial area of interest defined as a WKT Polygon or Point.
```
rod-glover: Ditto `stats`

doc/source/workflow.md
corviday: I would love to! I haven't figured out how yet. [It's not just me](https://github.com/miyakogi/m2r/issues/41).

doc/source/workflow.md
corviday: I would love to! I haven't figured out how yet. [It's not just me](https://github.com/miyakogi/m2r/issues/41).

doc/source/workflow.md
corviday: I would love to! I haven't figured out how yet. [It's not just me](https://github.com/miyakogi/m2r/issues/41).

doc/source/overview.md
corviday: Hm, good point.

doc/source/overview.md
corviday: I struggled writing the help with how much to try and stick  to generalities that applied to all possible data this API can serve versus how much to go into specifics on our most common case (specifically, GCM output climatologies containing a single variable of climate data). Most of our data, by far, is climatological data derived from a GCM and has an RIP code, but not all of it. So I wasn't sure whether to talk more about RIP codes or RCP codes, which apply to almost all, but not all, of the data. 

Thoughts or guidelines on generality versus detail?

rod-glover: I think you are right to want to discuss the most general case. However, the term "experiment" at minimum comes from a specific context and usage, and I think that's worth explaining, otherwise it just becomes unthethered jargon. To generalize, I'd suggest you say, "we generalized from this very common case arising from GCMs to ...."

rod-glover: Also, sorry about the single comments, I meant to submit a review as a single swack. I can let you know when I'm done so you don't jump every time I save a comment.

doc/source/workflow.md
```diff
@@ -0,0 +1,29 @@
...
+
+The `stats` endpoint returns statistical measures (`mean`, `stdev`, `min`, `max`, and `median`) of a single dataset identified by its unique identification string, a temporal index, and optionally a spatial area of interest defined as a WKT Polygon or Point.  
+
+You may wish to compare the statistical measure of this particular climatological mean datafile against the statistical measure of climatological means of other timespans to observe change over time. For example, if the file of interest is a climatological mean across the years 1961-1990, there may also be datafiles representing the same data for 1971-2000, 1981-2010, or other spans available. The `multistats` query may be called with parameters that describe a set of datasets by specifying all parameters except the start and end dates, as well as a time index and optional spatial area of interest. It responds with the same information as the `stats` query, but for every datafile that matches the parameters.
```
rod-glover: I think the description here may be just a little off. I'm not sure, but I think that the statistics describe the underlying dataset from which the climatological mean is computed. It looks, from your discussion in the previous paragraph, that this is for a fixed point in time, and so that these statistics are essentially spatial (i.e., mean, stdev, etc. over the spatial area of interest). The climatological statistics (e.g., mean) describe the dataset *temporally*, at all locations. In any case, my uncertainty here means that this documentation needs to be more explicit and precise.

rod-glover: On further thought, reading what's below, I think that `stats` is spatio-temporal; the temporal index identifies what time period, and the spatial argument identifies what spatial extent. 

This definitiely needs more precise explaining. Possibly it doesn't all belong here, but this does need to be made clear somewhere.

doc/source/api/api-overview.md
```diff
@@ -0,0 +1,4 @@
+Documentation for each API endpoint is automatically generated from the code and docstring for that API's main function and may not be entirely user-friendly. There are some very minor differences between arguments for each API function and the parameters needed for a web query.
+
+1. Web queries do not supply a `sesh` (database session) as an argument; that will be automatically done by the query parser.
+2. Parameters supplied in a query url should be web-encoded.
```
rod-glover: Let's be a little more explicit about the relationship between the function and the endpoint. Specifically,
- function parameters (except `sesh`) --> URL query parameters
- function value returned --> converted to JSON as endpoint response

doc/source/workflow.md
rod-glover: The `stats` API documentation is somewhat clearer on this point, but I think both it and this discussion could be made clearer. This one in particular.

doc/source/api/watershed-api-usage.md
```diff
@@ -0,0 +1,14 @@
+This API endpoint provides contextual information about the watershed that drains to the point specified by the parameters. It is intended to clarify and provide context for data about streamflow in watershed.
+
+Every grid cell is defined as flowing into a single other grid cell, so this data is most reliable for larger watersheds representing at least ten grid cells, and completely inappropriate for creeks or culverts smaller than a single grid cell. At small scales, streamflow variations within grid cells, not capturable by a gridded dataset, play too large a role.
+
+## Hypometric curve
```
rod-glover: "Hypsometric" (sp.)

doc/source/api/watershed-api-usage.md
```diff
@@ -0,0 +1,14 @@
...
+The `hypsometric_curve` object defines a histogram of area by elevation. 
+
+* Elevation bins are of equal width, `w = elevation_bin_width`.
+* Elevation bin `k` is bounded by elevations `ek` and `ek+1`, where `ek = e0 + (k * w)` for `0 <= k < n`
+* where `e0 = elevation_bin_start` and `n = elevation_num_bins`
+* The sum of areas of all cells with elevation falling into elevation bin `k` is given by `ak = cumulative_areas[k]`.
+* Units of elevation and area are specified by the properties `elevation_units` and `area_units`.
+
+Because elevation data is given as the mean of each grid square it is possible to have nonzero values for two elevation bins `ek` and `ek+2` but a zero value for `ek+1` in steep landscapes.
```
rod-glover: Wot, no LaTeX? Also, I have a fuller description of the problem for steep landscapes. "Source code" below:

```
# Hypsometric curve

The `/streamflow/watershed` API endpoint response includes a hypsometric curve for the watershed that drains to the specified mouth point. This curve is defined by the data in the `"hypsometric_curve"` property of the response JSON.

## Content

The `"hypsometric_curve"` property has the following format:

```json
"hypsometric_curve": {
    "elevation_bin_start": 0,
    "elevation_bin_width": 100,
    "elevation_num_bins": 46,
    "cumulative_areas": [
        // 46 numerical values
    ],
    "elevation_units": "m",
    "area_units": "m2"
}
```

This object defines a histogram of area by elevation.

Elevation bins are of equal width, $w =$ `elevation_bin_width`. 

Elevation bin $b$ is bounded by elevations $e_b$ and $e_{b+1}$, where 

$$
e_b = e_0 + bw \textnormal{ for } 0 \leq b < n
$$

where $e_0 =$ `elevation_bin_start` and $n =$ `elevation_num_bins`

The sum of areas of all cells with elevation falling in to elevation bin $b$ is given by $a_b =$ `cumulative_areas[b]`.

Units of elevation and area are specified by the properties `"elevation_units"` and `"area_units"`.



## Notes

### Gaps in hypsometric curve, or "empty" bins

For large areas of the earth and reasonably large elevation bins, we expect to see non-zero cumulative areas for each elevation bin between the minimum and maximum elevation over that area. In other words, there should be at least some area at each elevation in the histogram.

However, for small areas with steep topography, it is common to see some of the elevation bins between min and max elevation with zero area. This is not an error in either the computation or the data that feeds it. It is instead a product of the fact that $n$ surface grid cells can represent at most $n$ elevations. 

Consider the most extreme case of $n = 2$ cells that happen to be positioned at the edge of a steep sided valley. One cell is in the valley bottom with an average elevation of 100 m. The other is cell, just adjacent to it, mostly covers the highland above with an average elevation of 500 m. In a histogram with 100 m bin width, we'd see non-zero areas for the 100 m bin and the 500 m bin, but zero areas for the 200 m, 300 m, and 400 m elevation bins, and in the graph these would look like gaps. 

We can see a similar effect for other small values of $n > 2$ in steep terrain too. Once $n$ becomes large enough, then the likelihood of an elevation bin not having *some* cells is quite low and these gaps do not appear.
```

corviday: I have been able to figure out how to export _to_ LaTeX, but not how to render _from_ LaTeX. Though I am not particularly confident I have exhausted all the plugin options yet. 

Technical problems I want to solve and have reasonably hopes they're solvable:
- add links to markdown files that are correctly converted to point to the html files when the output is rendered
- easier build like the pdp has, and testing the buid
- render LaTeX

doc/source/overview.md
rod-glover: Yeah, I'd really prefer that. But if you think it too strained, it's not that big a deal.

doc/source/overview.md
rod-glover: And by that I meant _climatological_ statistics of those various kinds.

corviday:
@nikola-rados , could you look over this when you have time?

* Try downloading the branch and see if the documentation builds correctly (instructions have been added to the README).

* If documentation builds correctly, it will show up in doc/build/html. Read the documentation and see if it all seems correct, generally makes sense.

* Let me know if you think of anything else I should cover?

corviday:
- [ ]  add to the documentation documentation that you need to install `latexmk` to generate a PDF

PR #120, opened 2020-02-07
Allow users to skip calculations they don't require

### LINKED ISSUE #119: Allow callers to stats and multistats API to specify which statistics they are interested in
Currently the `stats` API returns a JSON object with attributes representing `min`, `max`, `mean`, `median`, `stdev`, and `ncells` values. Plan2adapt's calculations have been running slowly, and it might speed things up a bit if, when plan2adapt requested statistics, it could request only the exact statistics it needs.

Add parameter(s) to control which statistics are calculated.

corviday:
This PR adds optional parameters to the `multistats` and `stats` queries to allow a user to request a subset of the statistics normally provided by these queries. This goal of this change is to allow a minor performance improvement by skipping unnecessary calculations (https://github.com/pacificclimate/p2a-rule-engine/issues/3), but since most of the execution time is spent reading the datafile, improvements via this method are limited.

The optional parameters are named `median`, `mean`, `max`, `min`, and `stdev`, and are passed without arguments, for example:

```
http://server:port/api/multistats&ensemble_name=ce_files&model=PCIC12&variable=cdd
&emission=historical,rcp85&time=2&timescale=seasonal&mean&median
```
would return only the mean and median values, skipping calculation and return of `max`, `min` and `stdev`.

`ncells`, `units`, `time`, and `modtime` are always returned.

A demo docker is [here](http://docker-dev01.pcic.uvic.ca:30033/api/multistats?ensemble_name=ce_files&model=PCIC12&variable=cdd&emission=historical,rcp85&time=2&timescale=seasonal&mean).

Resolves #119

### DIFF
--- a/ce/tests/api/test_stats.py
+++ b/ce/tests/api/test_stats.py
@@ -52,3 +52,33 @@ def test_stats_bad_params(populateddb, unique_id, var):
 def test_stats_bad_id(populateddb):
     rv = stats(populateddb.session, 'id-does-not-exist', None, None, None)
     assert rv == {}
+
+@pytest.mark.parametrize('mean', ['', None])
+@pytest.mark.parametrize('median', ['', None])
+@pytest.mark.parametrize('max', ['', None])
+def test_stats_subset(populateddb, mean, median, max):
+    sesh = populateddb.session
+    unique_id = 'tasmax_mClim_BNU-ESM_historical_r1i1p1_19650101-19701230'
+    
+    #default all-statistics request for comparison
+    allstats = stats(sesh, unique_id, None, None, 'tasmax')
+    
+    #subset request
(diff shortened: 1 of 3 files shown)


ce/tests/api/test_stats.py
```diff
@@ -52,3 +52,33 @@ def test_stats_bad_params(populateddb, unique_id, var):
...
+    allstats = stats(sesh, unique_id, None, None, 'tasmax')
+    
+    #subset request
+    subset = stats(sesh, unique_id, None, None, 'tasmax', mean=mean, median=median, max=max)
+    
+    if mean is None and median is None and max is None:
+        assert allstats == subset
+    else:
+        for m in ['mean', 'median', 'max']:
+            if eval(m) is None:
```
rod-glover: Whoah

corviday:
This branch was part of the attempt to run the rules engine faster. It did help with that, but ultimately we went another route.

Issue #133, opened 2020-04-23
Render `percentileanomaly` data items as lists rather than dicts


rod-glover:
At present, `percentileanomaly` renders the data results (the anomaly and baseline data proper) as nested dicts, which are harder to filter than they need be. 

Instead of this:

```
{
  ...,
  "anomaly": {
    "yearly": {
      "2055-07-02 00:00:00": [ 2.3264379226208, 3.2458067672088 ]
    },
    "monthly": {
      "2055-01-15 00:00:00": [ 2.1310248585852, 3.3104322776824],
      "2055-02-15 00:00:00": [ 2.0116830378, 2.802655705519 ],
      "2055-03-15 00:00:00": [ 2.176002151962, 2.8835552547429 ],
      ...
    },
    "seasonal": {
      "2055-01-16 00:00:00": [ 2.7983318484797, 3.3868735586218 ],
      "2055-04-16 00:00:00": [ 2.3564425042040, 3.200660500285 ],
      "2055-07-16 00:00:00": [ 2.0237840492108, 3.1990939413123 ],
      ...
    }
  },
 ...
}
```
render this:

```
{

  ...,
  
  "anomaly": [
    // january; timeindex = 0
    {
      "timescale": "monthly",
      "date": "2055-01-15 00:00:00",
      "values": [ 2.1310248585852, 3.3104322776824]
    },
    
    ...,
    
    // winter; timeindex = 12
    {
      "timescale": "seasonal",
      "date": "2055-01-16 00:00:00",
      "values": [ 2.7983318484797, 3.3868735586218 ]
    },
    
    ...,
    
    // yearly; timeindex = 16
    {
      "timescale": "yearly",
      "date": "2055-07-02 00:00:00",
      "values": [ 2.3264379226208, 3.2458067672088 ]
    }
  ],
  
  ...

}
```

Notes:
1. Order matters: in order of "timeindex" as we call it in various places, i.e., jan, feb, mar, ..., winter, spring, summer, fall, yearly.
1. Key "timescale" in each data object is not strictly necessary but may be convenient. And would handle cases where various timescales were absent. For further contemplation.
1. No comments of course in the rendered JSON.

jameshiebert:
@rod-glover This should be fairly simple to implement, but just to confirm: this would be a non-backward-compatible API change and require a major version increment?

rod-glover:
Yes; only P2A is a client, and that code is recent, but this is a major version increment.

PR #137, opened 2020-05-12
percentileanomaly renders data object as a list

sum1lim:
As described in issue [#133](https://github.com/pacificclimate/climate-explorer-backend/issues/133), `percentileanomaly` in the master branch renders anomaly/baseline data as nested dictionaries:
```
{
  ...,
  "anomaly": {
    "yearly": {
      "2055-07-02 00:00:00": [ 2.3264379226208, 3.2458067672088 ]
    },
    "monthly": {
      "2055-01-15 00:00:00": [ 2.1310248585852, 3.3104322776824],
      "2055-02-15 00:00:00": [ 2.0116830378, 2.802655705519 ],
      "2055-03-15 00:00:00": [ 2.176002151962, 2.8835552547429 ],
      ...
    },
    "seasonal": {
      "2055-01-16 00:00:00": [ 2.7983318484797, 3.3868735586218 ],
      "2055-04-16 00:00:00": [ 2.3564425042040, 3.200660500285 ],
      "2055-07-16 00:00:00": [ 2.0237840492108, 3.1990939413123 ],
      ...
    }
  },
 ...
}
```
The following structure was desired for easier filtering process:
```
{

  ...,
  
  "anomaly": [
    // january; timeindex = 0
    {
      "timescale": "monthly",
      "date": "2055-01-15 00:00:00",
      "values": [ 2.1310248585852, 3.3104322776824]
    },
    
    ...,
    
    // winter; timeindex = 12
    {
      "timescale": "seasonal",
      "date": "2055-01-16 00:00:00",
      "values": [ 2.7983318484797, 3.3868735586218 ]
    },
    
    ...,
    
    // yearly; timeindex = 16
    {
      "timescale": "yearly",
      "date": "2055-07-02 00:00:00",
      "values": [ 2.3264379226208, 3.2458067672088 ]
    }
  ],
  
  ...

}
```

To achieve this goal, a helper function `convert_to_li(dict)` is added to `percentileanomaly`. This function takes a dictionary(in this case, it is nested dictionaries that holds anomaly/baseline data), and returns a list with multiple data objects(in a dictionary format) in it. The nested `for` loop iterates through the data stored in the nested dictionaries and converts them into data objects. When creating the data objects, an attribute `dateidx` is added for sorting. Once the data objects are sorted, `dateidx` is no loger needed, thus the function deletes the attribute from all of them.

The function is called for both anomaly and baseline data and is placed right before the declaration of `response`, which is the output for `percentileanomaly`. 

To test the new structure, `ce/tests/api/test_percentileanomaly.py` is also modified to have new `default_response`, which should represent the new desired output of percentileanomaly. Additionally, `def test_percentile_calculation(populateddb, num_percentiles)` also changed to perform a test on the list data structure.


ce/api/percentileanomaly.py
```diff
@@ -240,6 +240,34 @@ def canonical_timestamp(timestamp, timescale, timeidx):
                 projected_data[timescale][timestamp] = list(
                     np.percentile(anomalies, percentiles))
 
+        # converts a nested dictionary to a list storing data objects
+        def convert_to_li(dict):
```
jameshiebert, line 244: It's good that you set this functionality in a function!

I do think that you might be able to do it much more simply. You're on the right track with the `sort`ing using a function for the key. Did you know that your key function can return a tuple and Python with sort according to the first element of the tuple and then the next?

So you could combine a primary key as such: `lambda x: {"annual": 3, "seasonal": 2, "monthly": 1}[x]` with a secondary key that sorts on date. Then the whole function pretty much boils down to:

```python3
def convert_dict_to_sorted_list(d):
    ret_li = list(d)
    # define sort function
    ret_li.sort(key=sort_function)
    return ret_li
```

ce/api/percentileanomaly.py
```diff
@@ -240,6 +240,34 @@ def canonical_timestamp(timestamp, timescale, timeidx):
                 projected_data[timescale][timestamp] = list(
                     np.percentile(anomalies, percentiles))
 
+        # converts a nested dictionary to a list storing data objects
```
jameshiebert, line 243: Good comment to describe what the function does. It would be even better if you used this comment as [the function's docstring](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings).

ce/api/percentileanomaly.py
jameshiebert: > In particular, in the main loop starting on line 162, you already have the timeidx value direct from the rows retrieved from the CSV file

Ah, or use that! Even simpler.

corviday:
> In particular, in the main loop starting on line 162, you already have the `timeidx` value direct from the rows retrieved from the CSV file, and that would tell you where to place the rest of the values computed there in a list (as opposed to in the nested dictionaries).

The `timeidx` values in the CSV file are not the same as the `timeidx` values in the response specification and can't be used without some extra logic. The `timeidx` values in the CSV file correspond to the timestamp within the netCDF dataset files, so annual datasets have timeidx 0 (annual), seasonal datasets have available timeidx values of 0 (winter), 1 (spring), 2 (summer), 3 (autumn), monthly datasets have 0 (January), 1 (February), 2 (March), etc. A `timeidx` of 0 can mean either annual, winter, or January. 

(This is because the CSV file is stored results from the `/stats` query, so it lists all the arguments used the call the `/stats` query, and the `/stats` query uses `timeidx` as an argument.)

The response specification wants 0-16 covering all months, seasons, and annual.

rod-glover:
It seems we have a difference of opinion amongst reviewers here. In particular, @jameshiebert  is approving the dict-to-list approach. (Implicitly, @corviday may agree.) My opinion is that building the dict and then unravelling it leaves us with complicated, hard to understand code. However, I do acknowledge that reworking this to generate lists directly is a longer undertaking than converting the ready-made dicts. Does anyone have a comment about this?

jameshiebert:
Yeah, my way might not even work right.

```python3
>>> list({'foo': 1, 'bar': 2})
['foo', 'bar']
```

Listen to Rod. Also, never take two compasses to sea. :smile:

sum1lim:
Thanks for the feedback from all of you. I am working on reconstructing the function based on @rod-glover 's opinion and hopefully I can push the newer version of it before tomorrow's meeting.

ce/api/percentileanomaly.py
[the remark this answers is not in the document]
sum1lim: I thought about that too, but I figured `def add_to_nested_dict(dict, value, *keys):` uses `dict` as an argument, and I wanted to be consistent about it. For the newer version, should I change all the variables that are named `dict` in the file?

sum1lim:
@rod-glover I have made my second commit, which eliminated all the dictionary code. But as I metioned in the meeting today, I was afraid there might be a possible aggravation in terms of Big-Oh efficiency, since it is now using list data structure instead of dictionary data structure. The parts of code I am worried are the followings:
`line 208`
```
    def merge_duplicates(li):
        '''    
        this function accepts a list of multiple data objects and merges objects with the same dateidx.
        it is intended to reduce the size of the list and organize the data effectively for further process.
        '''
        i = 0
        while i < len(li)-1:
            if li[i]["dateidx"] == li[i+1]["dateidx"]:
                if(len(li[i]) == 5):
                    li[i]["values"] = li[i]["values"] + li[i+1]["values"]
                    li[i]["models"] = li[i]["models"] + li[i+1]["models"]
                    del(li[i+1])
                else:
                    li[i]["values"] = li[i]["values"] + li[i+1]["values"]
                    del(li[i+1])
            else:
                del(li[i]["dateidx"])
                i+=1
        if len(li) > 0:
            del(li[i]["dateidx"])
        return li
```

`line 308`
```
                for b in baseline_data:
                    baseline_date = datetime.strptime(b["date"], date_format)
                    if p["timescale"] == b["timescale"] and baseline_date.month == projected_date.month:
                        if not baseline:
                            b["values"] =b["values"][0]
                            baseline = float(b["values"])
                        else:
                            abort(500,
                                    ("Multiple matching baseline datasets ",
                                    "for {} {} {} {}").format(
                                        baseline_model,
                                        baseline_climatology,
                                        p["timescale"], p["date"]))      
```

I have no idea how big the input data can be, but I believe it should be fine since both parts are iterating lists with a fairly small amount of data objects.

ce/api/percentileanomaly.py
```diff
@@ -183,89 +279,69 @@ def canonical_timestamp(timestamp, timescale, timeidx):
...
-                                       "for {} {} {} {}").format(
-                                            baseline_model,
-                                            baseline_climatology,
-                                            timescale, timestamp))
-                    if baseline is None:
-                        abort(500,
-                              ("No baseline match available for ",
-                               "{} {} {} {}").format(
-                                   baseline_model, baseline_climatology,
-                                   timescale, timestamp))
```
sum1lim, line 225: I had to omit `abort(500, "Missing baseline data: {} {} {}".format(baseline_model, baseline_climatology, timescale))` error message since there is no way to check if the `timescale` is in `baseline_data` before the `for` loop. I don't see any problem not having this message as this error is a subset of the latter error:
```
                        abort(500,
                              ("No baseline match available for ",
                               "{} {} {} {}").format(
                                   baseline_model, baseline_climatology,
                                   timescale, timestamp))
```

rod-glover:
Sangwon, your general approach here is quite good. It has a a functional programming flavour (despite the abundance of objects that are mutated), in that you generate a list full of objects, then process that (via your `merge` function) into a better representation, then sort that. This was in fact the first thing that occurred to me, too, and it has some strengths.

In the service of learning, let's think about alternatives, however. This is always worth doing in a complicated situation (which is what you inherited here): Before you commit to a full solution, outline at least 2 (ideally 3 if you can come up with them) alternatives and give each a little thought: clarity and complexity of code, efficiency in execution. Partial code plus comments indicating what needs to be filled in are good for this kind of thinking.

It's also always worth starting with your desired output and thinking about how you can build it up most simply and directly. 
- The goal is ultimately to build a list of objects indexed by `dateidx`. 
- How long can that list be? At most 17: 12 months + 4 seasons + 1 annual.
- You start with some data (variable `queries`) that might present the information for building those objects in arbitrary `dateidx` order; and they might contain data that specifies more than one item per `dateidx`.

So, an alternative way we might approach this problem might be:
- Create an "empty" results list indexed by `dateidx`. Each element of the results list is itself a list of the data objects. Each of these sub-lists starts out empty, of course, because you haven't processed any data yet.
- Process the data (`queries`), computing `dateidx` and the corresponding derived data object, and fill up the results list by targeting `result[dateidx]`, and appending each data object 
- This would spare you the effort and complexity of going back over the list and merging the duplicate items it the way you did. In some sense you generate that merged list directly. 

A third alternative we might consider is to retain the method you use, but instead of merging your initial list in place, merge it into a separate results list. That retains the functional, list-processing flavour of your original solution, but simplifies the code, since you are not constantly mutating the input array.

Why don't you respond with:
- any other alternatives you can think of (they need not be wholly different than the 3 we now have)
- a quick evaluation of the strengths and weaknesses of the 3 (clarity, simplicity, efficiency)
- code sketches for at least one alternative (whichever you like better) ... feel free to use lots of comments to omit details and just give the outline of the code

rod-glover:
One more thing: efficiency is deliberately last on the list of evaluation criteria. We CS types tend to optimize prematurely, and often (as you suggest in your remarks) it doesn't matter. Clarity and maintainability are often *far* more important than efficiency. 

That said, I do really appreciate your attention to big-O complexity of your solutions, since that is the kind of efficiency that a good optimizing compiler cannot create for us!

sum1lim:
@rod-glover 
```
A third alternative we might consider is to retain the method you use, but instead of merging the initial list in place, merge it into a separate results list. That retains the functional, list-processing flavour of the original solution, but simplifies the code, since we are not constantly mutating the input array.
```
I find this alternative is not very intuitive compared to another one you suggested(maybe I am not fully understanding the concept). I do not see the necessity of keeping the merging process and creating a separate list looks redundant.

However, I like your first suggestion. But instead of starting with empty sublists, I came out with an idea that initializes data objects as empty dictionaries.

Idea 1:
```
def add_to_nested_li(li, value, timeidx):
	#determine dateidx depending on the timescale
	if monthly:
		dateidx = timeidx
	elif seasonal: 
		dateidx = timeidx + 12
	else
		dateidx = 16
	
if “values” in li[timeidx].keys():
	li[timeidx][“values”].append(value)
	else:
		#create a data_object
		obj = create_data_object(row["mean"], row["timescale"],
                                          			ctimestamp, row["model"])
		li[timeidx] = obj
```
```
# start with a list of 17 empty objects
projected_data = [{},{},...,{}]

	....

for row in queries:

	....

	add_to_nested_li(projected_data, obj, row[“timeidx”])
```

Strengths
The code is clearer since the data objects are organized based on the `dateidx`
The code is more intuitive than having the complex merging process
The list is sorted since its creation, thus no further sorting process required
Does not require merging function
Only have to iterate `queries` once, while merging process required reiteration of it
Data_objects does not need to carry `dateidx` attribute throughout the rest of the program, thus no need to be removed as well

Weaknesses
The list may include unnecessary empty objects, therefore they need to be omitted after.

rod-glover:
Excellent idea and analysis.

rod-glover:
Don't worry about creating a list with empty objects that may need to be removed. This is a minor efficiency concern whose "fix" would make the code much more complex. In general, it is better to write code that looks like:

```
#  simple code that creates a list `x` with some null/empty/undesirable values
y = compress(x)
```

than something like:

```
# complicated code that somehow avoids producing null/empty/undesirable values
``` 

We've all been misled by the idea that storage is at a premium. It's not. Feel free to create and destroy lists at will. The garbage collector is on your side. Big memory is on your side.

rod-glover:
I'd also suggest that we consider `None` as the element values in the initial list. It's not necessary to use empty objects there, since you replace them anyway, and it communicates a little more firmly that there is nothing there. 

In case you are not aware of them, there is a very nice way to do this in Python. It uses what is called a "list comprehension" and also a handy built-in function `range`:

```
projected_data = [None for i in range(17)]
```

Definitely read up on both if you are not familiar with them. Knowing these are essential Python skills.

rod-glover:
As I said, very nice work so far!

ce/api/percentileanomaly.py
```diff
@@ -183,62 +291,66 @@ def canonical_timestamp(timestamp, timescale, timeidx):
...
-                                           row['timescale'], ctimestamp)
+                        obj = create_data_object(row["mean"], row["timescale"], ctimestamp)
+                        idx = generate_list_idx(row["timescale"], row["timeidx"])
+                        baseline_data[idx] = obj
+
+
+        projected_data = [data_object for data_object in projected_data if data_object != None]
+        baseline_data = [data_object for data_object in baseline_data if data_object != None]
 
         # calculate percentiles and anomalies
```
rod-glover, line 308: This chunk of code (the entire for loop) is long and hard to understand. I think we might do better to break out the part that determines the baseline as a separate function that is called inside the loop, leaving the remainder as-is. It would also help my tiny brain if there were some comments explaining the baseline calculation a bit, specifically the conditions under which a baseline found. These would naturally go in the baseline function.

ce/api/percentileanomaly.py
```diff
@@ -183,62 +291,66 @@ def canonical_timestamp(timestamp, timescale, timeidx):
...
+                            p["timescale"], p["date"], p["models"])
+            elif values.size > 12:
+                abort(500,
+                        "Extraneous data for {} {}. Models available: {}").format(
+                            p["timescale"], p["date"], p["models"])
+
+            anomalies = values - baseline
+
+            p["values"] = list(np.percentile(anomalies, percentiles))
+            del(p["models"])
```
rod-glover, line 350: We do this because we don't want the "models" key in the final result?

ce/tests/api/test_percentileanomaly.py
```diff
@@ -125,8 +252,7 @@ def test_percentile_calculation(populateddb, num_percentiles):
...
-        for timestamp in response["anomaly"][resolution]:
-            percs = response["anomaly"][resolution][timestamp]
-            assert len(percs) == num_percentiles
-            for i in range(num_percentiles-1):
-                assert percs[i] < percs[i+1]
\ No newline at end of file
+        percs = resolution["values"]
+        assert len(percs) == num_percentiles
+        for i in range(num_percentiles-1):
+            assert percs[i] < percs[i+1]
```
rod-glover, line 258: :+1:

ce/api/percentileanomaly.py
sum1lim: Yes, it is. I did it because the original `percentilanomaly` outputs the nested dictionaries without having the `models` attribute in the very inner loop. Should I keep it to have the attribute?

ce/api/percentileanomaly.py
```diff
@@ -292,64 +298,89 @@ def canonical_timestamp(timestamp, timescale, timeidx):
...
+                                baseline_model,
+                                baseline_climatology,
+                                p_obj["timescale"], p_obj["date"]))
+    
+            return baseline
+
+
+        # calculate percentiles and anomalies
+        curr_idx = 0
+        while curr_idx < len(projected_data):
```
rod-glover, line 347: What we are doing here with `curr_idx` is a very common pattern, and Python provides a nicer tool for it than the `while` loop: We can (and should) use [`enumerate` ](https://docs.python.org/3.8/library/functions.html#enumerate) here. It's simpler and it's way harder to make a mistake with. Python is full of all kinds of goodies like this!

ce/api/percentileanomaly.py
```diff
@@ -292,64 +298,89 @@ def canonical_timestamp(timestamp, timescale, timeidx):
...
-            del(p["models"])
+        def remove_None(li):
+            '''
+            the function accepts a list and removes every None values in the list
+            '''
+            return [obj for obj in li if obj is not None]
 
 
+        projected_data = remove_None(projected_data)
+        baseline_data = remove_None(baseline_data)
```
rod-glover, line 383: Sweet! Now that is DRY as a bone :)

ce/api/percentileanomaly.py
```diff
@@ -183,62 +297,91 @@ def canonical_timestamp(timestamp, timescale, timeidx):
...
+
+                        projected_data, model_counter = add_to_nested_li(projected_data, row, ctimestamp, model_counter)
+
                     elif (calculate_anomaly and
                             row['model'] == baseline_model and
                             row['climatology'] == baseline_climatology):
-                        add_to_nested_dict(baseline_data, row['mean'],
-                                           row['timescale'], ctimestamp)
+
+                        baseline_data = add_to_nested_li(baseline_data, row, ctimestamp)
```
rod-glover, line 307: To receive the uniform result of this, use `baseline_data, _ = ...`. The `_` is a common idiom for "thing I don't care about."

ce/api/percentileanomaly.py
```diff
@@ -183,62 +297,91 @@ def canonical_timestamp(timestamp, timescale, timeidx):
...
+
+
+
+        def determine_baseline(curr_idx, p_obj, baseline_data, baseline_model, baseline_climatology):
```
rod-glover, line 311: Place this outside the loop, and for that matter, outside the `try`. It is (or at least should be) independent of any variables being maintained in the loop, and technically speaking you are redefining the function each time through the loop. You want to avoid that, mainly for clarity (the Python interpreter is pretty smart and will probably avoid actually recreating the function each time).

ce/api/percentileanomaly.py
```diff
@@ -292,64 +298,89 @@ def canonical_timestamp(timestamp, timescale, timeidx):
...
+
+            if p_obj is not None:
+                if calculate_anomaly:
+                    baseline = determine_baseline(curr_idx, p_obj, baseline_data, baseline_model, baseline_climatology)
+                else:
+                    baseline = 0.0
+
+                values = np.asarray([float(v) for v in p_obj["values"]])
+
+                num_models = len(model_counter[curr_idx])
```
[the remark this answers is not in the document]
sum1lim: @rod-glover I actually tried to check if there are any duplicates of models. But still, I do not know if there is any circumstance like that

ce/api/percentileanomaly.py
```diff
@@ -183,62 +326,54 @@ def canonical_timestamp(timestamp, timescale, timeidx):
...
+                            "Extraneous data for {} {}. Models available: {}").format(
+                                p_obj["timescale"], p_obj["date"], p_obj["models"])
+
                 anomalies = values - baseline
-                projected_data[timescale][timestamp] = list(
-                    np.percentile(anomalies, percentiles))
+
+                p_obj["values"] = list(np.percentile(anomalies, percentiles))
+
+            curr_idx += 1
```
rod-glover, line 364: Remove this.

ce/api/percentileanomaly.py
```diff
@@ -122,24 +176,74 @@ def percentileanomaly(sesh, region, climatology, variable, percentile='50',
...
+        Index 16 is assigned to "yearly" timescale.
+        '''
+        if timescale == "monthly":
+            idx = int(timeidx)
+        elif timescale == "seasonal": 
+            idx = int(timeidx) + 12
+        else:
+            idx = 16
+
+        return idx
```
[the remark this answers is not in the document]
rod-glover: Were you planning to do this?

ce/api/percentileanomaly.py
```diff
@@ -185,24 +185,19 @@ def generate_list_idx(timescale, timeidx):
...
-        elif timescale == "seasonal": 
-            idx = int(timeidx) + 12
-        else:
-            idx = 16
-
-        return idx
+            return int(timeidx)
+        if timescale == "seasonal": 
+            return int(timeidx) + 12
+        return 16
```
rod-glover, line 191: :+1:

rod-glover:
LGTM. We need to get the Jenkins issue clarified before we merge. And the tests have to pass.

jameshiebert:
Looks great! I'll merge away!

PR #139, opened 2020-05-12
Reduce image size

nikola-rados:
Uses slim base image to reduce overall size.

### DIFF
--- a/docker_requirements.txt
+++ b/docker_requirements.txt
@@ -1,0 +1,19 @@
+Flask
+Flask-SQLAlchemy
+Flask-Cors
+modelmeta==0.2.0
+shapely>=1.6
+numpy
+netcdf4==1.3
+rasterio~=1.0
+sqlalchemy==1.2.18
+contexttimer==0.3.3
+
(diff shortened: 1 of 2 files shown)


docker_requirements.txt
```diff
@@ -0,0 +1,19 @@
+Flask
```
jameshiebert, line 1: I think I see what you're trying to do here... install everything but GDAL since it's already installed in the base container?

But the down side is that this is almost 100% duplication of `requirements.txt`.

Maybe we should hold off on this. I already have [a branch](https://github.com/pacificclimate/climate-explorer-backend/tree/upgrade-geo) going that upgrades everything to GDAL 3.x and uses the pcic/geospatial-python:gdal3 image. It should be ready shortly...

nikola-rados: Sounds good!

jameshiebert:
Closing this in favor of the [`upgrade-geo` branch](https://github.com/pacificclimate/climate-explorer-backend/tree/upgrade-geo).

PR #141, opened 2020-05-13
Add PEP8 Styling

eyvorchuk:
This PR makes the 20 test files with violations PEP8 compliant.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,8 @@ before_install:
   - export GDAL_MINOR_VERSION=$(echo $GDALVERSION | sed -e "s/\\.[0-9]\+$//")
   - "pip install --global-option=build_ext --global-option='-I$GDALINST/include' --global-option='-L$GDALINST/lib' --global-option='-R$GDALINST/lib' GDAL==$GDAL_MINOR_VERSION"
 install:
-  - pip install -I pytest>=3.0.0
-  - pip install -r requirements.txt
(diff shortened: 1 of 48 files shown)


jameshiebert:
Did you notice above that the Jenkins tests are failing? Turns out that the Jenkins test run and the Travis test run were *slightly* different (Jenkins was being more thorough).

I've pushed a commit that brings the both in line. Review the Travis output... there are still another 22 source files that need to be `flake8`ified.

eyvorchuk:
The commit I just pushed fixes the Flake8 errors on the 22 source files in question, but `ce/tests/helpers/test_utils.py` still has an `E741 ambiguous variable name 'O'` error. I didn't do anything about this yet because `ce/api/streamflow/watershed.py` uses this variable.

jameshiebert:
Hmmm, I get that too, but it's weird that the test suite does not :shrug:

You can choose to address that if you would like, by coming up with an *unambiguous* variable name that can replace 'O'. `flake8` is right. That variable name is not great in that:

1. It's unclear if it's an Oh or a Zero
1. What does it mean?!

It's *probably* a good idea to fix that. Once you do, this looks ready to merge.

eyvorchuk:
Something like `Out` for outlet?

jameshiebert:
Why abbreviate? `OUTLET` is pretty short already :)

PR #145, opened 2020-05-27
Add Github Actions

nikola-rados:
Adds github actions to the repo.  The workflows will take over the responsibilities of Jenkins and Travis which are:
- run python test suite (on all pushes)
- build and publish docker image (on all branches + on tag creation)
- anchore scan docker image (on PRs targeting `master`)

I do not have permission to access the webhooks in this repo but Travis and Jenkins need to be deactivated.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -6,7 +6,7 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
-    contianer: pcic/geospatial-python:gdal3
+    container: pcic/geospatial-python:gdal3
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -1,0 +1,24 @@
+name: CI Tests
+
+on: push
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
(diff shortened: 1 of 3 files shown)


corviday:
It seems to push a docker image before tests are complete. Is that the intention? It probably doesn't matter, but if we ever want to go to automatic deployment, it would mean a broken image could be deployed.

corviday:
```
WARNING! Your password will be stored unencrypted in /github/home/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store
```

Do we need to do something about this?

nikola-rados:
> It seems to push a docker image before tests are complete. Is that the intention?

That is a good point @corviday. I was hoping to save some CI time by having the image build while the tests are running.  We can change it to wait for the tests to succeed in the future when we get to automated deployment.

> Do we need to do something about this?

Huh, I have not seen that before.  I'll take a look and see if there is something to be done about it.

corviday:
Maybe it doesn't matter because nobody is going to have access to the temporary container being run just for tests?

corviday:
I didn't test any tag functionality, but other than that it all seems to be working, aside from the ominous message about unencrypted passwords in the docker image publication.

nikola-rados:
I think we are okay.  Even using the official docker action produces the same warning.  Like you said it is being run in a container that folds once its been finished.  Furthermore there are no artifacts left behind from the run.

nikola-rados:
@corviday do you have permissions for this repo to deactivate the webhooks?

corviday:
Afraid not.

Issue #97, opened 2019-04-11
Odd behaviour in make_mask_grid()


corviday:
`make_mask_grid()` [generates a unique key for a combination of map grid and polygon](https://github.com/pacificclimate/climate-explorer-backend/blob/master/ce/api/geo.py#L137); it's used to determine whether a new polygon request has been previously made and doesn't need to be recalculated. The key consists of a tuple with:

- minimum longitude of the grid (float)
- maximum longitude of the grid (float)
- number of longitude steps (float)
- minimum latitude of the grid (float)
- maximum latittude of the grid (float)
- number of latitude steps (float)
- text (wkt) representation of the polygon to be computed (string)

However, during testing today, @nikola-rados was seeing singleton numpy arrays instead of numpy floats for the minimums and maximums, which is not the intended or previously observed behaviour. The obvious cause would seem to be a numpy update - numpy is not pinned in this repository - but we tried testing the steps in the python console with the newest version of numpy and they behaved as I would have expected.

This issue can be easily fixed by switching to `np.min()` instead of `np_array[0] `to determine the minimum of each coordinate, but I'd like to know _why_ it's happening and if there are any other side effects first, rather than blindly patching it. Needs further investigation.

corviday:
This issue doesn't show up running the tests, so there's apparently something specific about the way plan2adapt is calling polygonal queries that causes it, rather than some kind of update.

jameshiebert:
Here's sort of a minimal demonstration of this behavior, at the lowest level.

```python
$ python
Python 3.6.4 (default, Mar 20 2018, 13:12:40) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> import numpy.ma
>>> a = numpy.ma.masked_array([1, 2, 3, 4, 5], mask=[True, False, False, False, False])
>>> a[0]
masked
>>> key = a[0]
>>> value = 'value'
>>> {key: value}
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'MaskedConstant'
>>> key = a[1]
>>> {key: value}
{2: 'value'}
>>> numpy.min(a)
2
```

Not sure why lat/lon would be masked arrays, but maybe this gives you a lead?

corviday:
The issues seems to be from netCDF4 version 1.5.0.1. 

[...]

PR #147, opened 2020-07-02
Upgrade python-netCDF4 requirement to 1.5

jameshiebert:
The `python-netCDF4` library has released a couple of minor versions, and I think it's time to circle back on the decision that we made in #97 

This PR:

+ adds a test for our `geo` module that ensures that the array results that come back from a netCDF file are in fact scalars and hashable (which I believe was the original problem for the P2A rules engine)
+ upgrades `netCDF4` to 1.5 to demostrate the tests (expectedly) *failing*, and then
+ implements the fix

You can see the results form each step [in the Actions output](https://github.com/pacificclimate/climate-explorer-backend/actions?query=branch%3Aupgrade-netCDF4).

I'm pretty sure that it works well, but I'm happy to have another couple pair of eyes on this, just in case.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -12,6 +12,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: Install dependencies
       run: |
+        pip install -U pip wheel
         pip install -i https://pypi.pacificclimate.org/simple -r requirements.txt -r test_requirements.txt
         pip install .
         python setup.py build_sphinx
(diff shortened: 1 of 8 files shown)


jameshiebert:
Thanks you two. I'm just tidying a few things up in the dependency chain (`ceb` -> `modelmeta` -> `nchelpers`), and then I'll merge. Thanks for taking a look.

jameshiebert:
@corviday I'm pretty sure that this testing failure is related to the upgrade to discrete geometries in modelmeta (i.e. upgraded from 0.2.0 to 1.0.0). Would you be able to look at these test failures and point me in the right direction as to how to resolve them?

jameshiebert:
Actually, I think I have it figured out. Stand by...

jameshiebert:
*Finally* got everything working! Merging away.

PR #159, opened 2020-08-07
Add pre-commit hook

### LINKED ISSUE #158: Black as a Git pre-commit hook
The differences between what I can make my IDE enforce and what Black demands is driving me up the wall. There is a solution, and it's one we've discussed before. It's called a pre-commit hook, which is (roughly) a script that runs before every commit. It's now easier than ever:

- Package [pre-commit](https://pre-commit.com/) makes installing pre-commit hooks very easy.
- [This blog post](https://www.mattlayman.com/blog/2018/python-code-black/) describes how to set up Black as a pre-commit hook. (Search for "Black as a Git pre-commit hook", well down the page.)

I'd be happy to give this a go. If/when we decide we like it, we can transfer it to other Python repos. @nikola-rados , do you have any comment?

nikola-rados:
Resolves #158.

## Changes
- Add `pre-commit` configuration file
- Update `README`

## Notes
`pre-commit` was a great idea @rod-glover! I'm really happy with the interaction on the terminal. It will be nice to prevent simple styling from breaking actions (although it is good we are still checking for it there). The `pip` package is not in either of the requirements files because it was breaking github actions.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,5 @@
+repos:
+-   repo: https://github.com/psf/black
+    rev: 19.10b0
+    hooks:
+    -   id: black
--- a/README.md
+++ b/README.md
@@ -63,6 +63,13 @@ $ source venv/bin/activate
 (venv)$ pip install -e .
 ```
 
+#### Pre-commit hook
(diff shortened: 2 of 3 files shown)


nikola-rados:
I added a line to the `README` for setup. So ensure you run that on you machine to get the hook going. I am in the process of automating project setups, so hopefully carefully reading the `README` for install instructions will soon be in the past.

PR #163, opened 2020-08-31
Add makefile

### LINKED ISSUE #160: Automate installation
There are enough steps to the installation process that it would benefit from automation.

nikola-rados:
# Changes
- add `makefile`
- update `README.md`

# Notes
`make` seems to be the best candidate for automating our project setup. There are still some things I would like to implement in the future (`gdal` installation, more sophisticated dependency management, etc..) but I wanted to get something in here to jump-start the process of adding in these `makefiles`. Furthermore, I wanted to get some feedback and ensure we are covering all the bases.

Resolves #160

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,52 @@
+.PHONY: all
+
+# Used for GDAL installation
+export CPLUS_INCLUDE_PATH=/usr/include/gdal
+export C_INCLUDE_PATH=/usr/include/gdal
+export PIP_INDEX_URL=https://pypi.pacificclimate.org/simple
+
+# Makefile Vars
+SHELL:=/bin/bash
+VENV?=~/tmp/ceb-venv
+PYTHON=${VENV}/bin/python3
+PIP=${VENV}/bin/pip
+
+all: apt gdal-check install build-docs pre-commit-hook test
+
+apt:
+	sudo apt-get install -y \
+		libpq-dev \
+		python3-dev \
+		libhdf5-dev \
+		libnetcdf-dev \
+		libgdal-dev
+
+build-docs: venv
+	${PYTHON} setup.py build_sphinx
+
+clean:
+	rm -rf $(VENV)
+
+gdal-check:
+	$(eval GDAL_REQ=$(shell cat requirements.txt | grep -i gdal | cut -c 7-))
+	$(eval GDAL_VER=$(shell gdal-config --version))
+
+	if ! [ "$(GDAL_REQ)" == "$(GDAL_VER)" ]; then \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,52 @@
+.PHONY: all
+
+# Used for GDAL installation
+export CPLUS_INCLUDE_PATH=/usr/include/gdal
+export C_INCLUDE_PATH=/usr/include/gdal
+export PIP_INDEX_URL=https://pypi.pacificclimate.org/simple
+
+# Makefile Vars
+SHELL:=/bin/bash
+VENV?=~/tmp/ceb-venv
```
jameshiebert, line 10: I still think that we should use the system's temp directory (so $TMPDIR if set, else /tmp), but it can't be random (i.e. we can't use the `mktemp` command because it has to persist from one run to the next).

Maybe we can make it:

${TMPDIR:-/tmp}/ceb-venv

Not sure if that will work exactly *in* the Makefile, but that's how you express it in `bash`.

rod-glover (commented):
This looks straightforward. Question: Why is the project being set up in `~/tmp` (or whatever) and not, say in just in a subdirectory of the repo directory?

Makefile
nikola-rados: Its not quite as pretty, but I did manage to set this up.

nikola-rados:
@rod-glover it was suggested by @jameshiebert to do it this way. I think it is a way to keep the clutter down and ensure the venvs we are working with don't accidentally fall out of date. I could be wrong though.

jameshiebert:
Because:
1. a venv will usually run faster if they're on `$TMPDIR` (`/tmp` is often on a RAMDISK). This is *especially* true if you're working on glusterfs.
1. venv's actually take up quite a bit of disk space when they're collected together. Especially if you have lots of projects on the go. Creating them in `$TMPDIR` ensures that they get cleaned up regularly and automatically.
1. Re-creating your venv frequently ensures that devs aren't missing anything in the requirements and the automated process... helps ensure that everything is coded.

nikola-rados:
Thanks for that overview! My `$TMPDIR` doesn't seem to be set to anything, is there supposed to be a default value/location?

rod-glover:
Ditto. Learned something useful from that!

jameshiebert:
> My $TMPDIR doesn't seem to be set to anything, is there supposed to be a default value/location?

It probably won't be on your workstation, but if you're on a system with, say, an SSD-backed scratch pad, then in might be. E.g.:

```bash
hiebert@lynx:~$ echo $TMPDIR
/local_temp/hiebert
```

```bash
[hiebert@cedar5 ~]$ echo $TMPDIR
/tmp
```

PR #174, opened 2020-12-04
Add Melton Ratio to watershed API

### LINKED ISSUE #172: Add relief ratio to watershed API
MOTI has requested the [relief ratio](https://en.wikipedia.org/wiki/Relief_ratio) be added to the `watershed` API. This is the difference between the elevation of the source of a stream and its mouth divided by the length of the stream, but it's not entirely clear what "Source" means. Arelia will consult with Markus and get back to use on what the exact algorithm should be. 

The `watershed` API has all the relevant data (elevation, flow direction map) already, we just need to figure out the algorithm.

corviday:
MOTI requested the addition of the Melton ratio to the `watershed` API, defined as the change in elevation over a watershed divided by the square root of the area of the watershed. 

As the `watershed` API already provided the maximum elevation, minimum elevation, and total area of a watershed, this is a straightforward addition.

Resolves #172


ce/api/streamflow/watershed.py
```diff
@@ -123,6 +123,8 @@ def worker(station_lonlat, flow_direction, elevation, area, hypso_params=None):
         raise ValueError("Flow direction and elevation do not have compatible grids")
     if not flow_direction.is_compatible(area):
         raise ValueError("Flow direction and area do not have compatible grids")
+    if elevation.units != "m" or area.units != "m2":
+        raise ValueError("Elevation and area do not have compatible units")
```
rod-glover, line 127: Quibble: Either the condition is too strict (only `m` as the length unit) or the error message is too general. For example, `yd` and `yd^2` are compatible but would raise this error.

rod-glover: My personal inclination would be to use a units-parsing utility and do a more general comparison. Unless `m` really is  the only acceptable unit. However, that's more effort.

rod-glover (approved):
LGTM, apart from a quibble.

ce/api/streamflow/watershed.py
corviday: That was my first line of attack, but `pint` doesn't understand `m2`!

rod-glover: Or pass run the string through a search/replace that inserts a `^` before digit strings.

corviday:
Added pint for unit checks; we'll also need to update the data used by this API, as at present it uses `m2` as the unit of area, which is not recognized by pint.

rod-glover:
This looks great! No fun fixing the units in the datases, but I think it is the cleanest solution.

PR #182, opened 2021-03-04
Add filtering on cell_methods to multimeta API

corviday:
This is a companion PR to #176, which added the ability to specify a cell_methods parameter (either "mean", "percentile", or "standard_deviation") to the `/data` API in order to support displaying datasets that are percentiles of a model ensemble. (The `/multistats` API already had this capability.)

This PR adds the same capacity to the `/multimeta` API.


rod-glover (approved):
LGTM!

PR #188, opened 2021-03-23
Allow filtering by Nth percentile in data,multimeta, and multistat

corviday:
This PR extends the `climatological_statistic` parameter, with which a user specifies whether they want data from climatological means, climatological standard deviations, or climatological percentiles.The new parameter is called `percentile` and allows a user to specify exactly _which_ percentile values they would like data or metadata from, 5th percentile, 50th percentile, etc. 

The new `percentile` parameter has been added to all the APIs where one selects multiple datasets by specifying a set of parameters that describe them: `multimeta`, `multistats`, and `data`. For `multistats` and `data`, the `percentile` parameter must be specified if the `climatological_statistics` is `percentile`. That is,  with these two APIs, you cannot merely filter to all percentile datasets, you must filter to a specific percentile value, such as 45th percentile datasets. For the `multimeta` API, the parameter is optional, you may request metadata on either 45th percentile datasets, or on all percentile datasets.

This change will allow the front end to select percentiles for graphing or display.

**Question for reviewers**
Three APIs, `multimeta`, `multistat`, and `data` currently support the `climatological_percentile` attribute. These APIs  have a caller describe the dataset(s) they are interested in as a collection of filtering parameters; the other APIs return data on a single datafile and require its `unique_id`.

I added the `percentile` parameter to all three of these APIs. My reasoning was that it was good to have the APIs have similar interfaces. However, I'm not sure about this decision and would like thee reviewer to weigh in. Here are the considerations:

1. a user can specify `climatological_statistic` as `percentile` to access data or metadata from all percentile datasets at once, if allowed. A user can additionally specify `percentile` as `5` to see only 5th percentile datasets.

1. the PCEX frontend needs the `multimeta` API to be able to return a list of all percentile datasets (this is how it learns what percentiles are available). It does not currently need the `multimeta` API to be able to provide a list of all 5th percentile datasets, but I could imagine this theoretically being useful sometime.

1. The PCEX frontend needs to be able to request a specific percentile from the `data` API. The `data` API cannot display data from multiple percentiles at once (doesn't fit in its "run" format).

1. The PCEX frontend does not need to be able to request either a specific percentiles or all percentiles from the `multistat` API, and furthermore, requesting standard deviation of a percentile dataset is kind of a mathematical mess and I don't think we'll ever want to do it.

Given that, which of the three APIs should I add the new `percentile` parameter to? `data` is the one we need. We could want `multimeta` someday. We'll probably never want it for `multistat` and I only put it there for consistency.

### DIFF
--- a/ce/api/data.py
+++ b/ce/api/data.py
@@ -26,6 +26,7 @@ def data(
     timescale="other",
     ensemble_name="ce_files",
     climatological_statistic="mean",
+    percentile=None,
     is_thredds=False,
 ):
     """Delegate for performing data lookups across climatological files
@@ -60,6 +61,10 @@ def data(
         climatological_statistic (str): Statistical operation applied to variable in a
             climatological dataset (e.g "mean", "standard_deviation",
             "percentile). Defaulted to "mean".
+        
+        percentile (float): if climatological_statistic is "percentile", specifies what
+            percentile value to use. A percentile value must be provided if the 
+            climatological_statistic is "percentile".
 
         is_thredds (bool): If set to `True` the filepath will be searched for
             on THREDDS server. This flag is not needed when running the backend
@@ -111,7 +116,6 @@ def data(
 
     """
     # Validate arguments
-
     try:
         time = int(time)
     except ValueError:
@@ -123,6 +127,21 @@ def data(
             "Unsupported climatological_statistic parameter: {}".format(
                 climatological_statistic
             )
+        )
+    if percentile is not None:
+        try:
+            percentile = float(percentile)
+        except ValueError:
+            raise Exception(
+                "Percentile parameter {} not convertable to a number".format(percentile)
+            )
+    if climatological_statistic == "percentile" and not percentile:
+        raise Exception(
+            "Percentile parameters must be specified to access percentile data"
+        )
+    if climatological_statistic != "percentile" and percentile:
+        raise Exception(
+            "Percentile parameter is only meaningful for percentile climatologies"
         )
 
     def get_spatially_averaged_data(data_file, time_idx, is_thredds):
@@ -184,7 +203,10 @@ def data(
         dfv
         for dfv in data_file_variables
         if check_climatological_statistic(
-            dfv.variable_cell_methods, climatological_statistic, True
+            dfv.variable_cell_methods,
+            climatological_statistic,
+            default_to_mean=True,
+            match_percentile=percentile,
         )
     ]
 
--- a/ce/api/data.py
+++ b/ce/api/data.py
@@ -112,18 +112,19 @@ def data(
             }
 
     Raises:
-        Exception: If `time` parameter cannot be converted to an integer
+        ValueError: If `time` or `percentile` parameters cannot be converted
+        to an integer, if climatological_statistic parameter is not recognized
 
     """
     # Validate arguments
     try:
         time = int(time)
     except ValueError:
-        raise Exception(
+        raise ValueError(
             'time parameter "{}" not convertable to an integer.'.format(time)
         )
     if not is_valid_clim_stat_param(climatological_statistic):
-        raise Exception(
+        raise ValueError(
             "Unsupported climatological_statistic parameter: {}".format(
                 climatological_statistic
             )
@@ -132,7 +133,7 @@ def data(
         try:
             percentile = float(percentile)
         except ValueError:
-            raise Exception(
+            raise ValueError(
                 "Percentile parameter {} not convertable to a number".format(percentile)
             )
     if climatological_statistic == "percentile" and not percentile:
--- a/ce/api/multistats.py
+++ b/ce/api/multistats.py
@@ -15,6 +15,7 @@ def multistats(
     variable="",
     timescale="",
     climatological_statistic="mean",
+    percentile=None,
     is_thredds=False,
 ):
     """Request and calculate statistics for multiple models or scenarios
@@ -53,6 +54,10 @@ def multistats(
         climatological_statistic (str): Statistical operation applied to variable in a
             climatological dataset (e.g "mean", "standard_deviation",
             "percentile"). Defaulted to "mean".
+            
+        percentile(float): the percentile value to filter on if the 
+            climatological_statistic is percentile. Required if climatological_statistic
+            is percentile.
 
         is_thredds (bool): If set to `True` the filepath will be searched for
             on THREDDS server. This flag is not needed when running the backend
@@ -95,6 +100,23 @@ def multistats(
             }
     """
 
+    # validate inputs
+    if percentile is not None:
+        try:
+            percentile = float(percentile)
+        except ValueError:
+            raise Exception(
+                "Percentile parameter {} not convertable to a number".format(percentile)
+            )
+    if climatological_statistic == "percentile" and not percentile:
+        raise Exception(
+            "Percentile parameters must be specified to access percentile data"
+        )
+    if climatological_statistic != "percentile" and percentile:
+        raise Exception(
+            "Percentile parameter is only meaningful for percentile climatologies"
+        )
+
     ids = search_for_unique_ids(
         sesh,
         ensemble_name,
@@ -104,5 +126,6 @@ def multistats(
         time,
         timescale,
         climatological_statistic,
+        percentile,
     )
     return {id_: stats(sesh, id_, time, area, variable, is_thredds)[id_] for id_ in ids}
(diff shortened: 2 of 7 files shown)


corviday:
Demo queries:

**/multimeta**
* [multimeta - metadata for all climatological means](http://docker-dev01.pcic.uvic.ca:30402/api/multimeta?ensemble_name=upper_fraser)
* [multimeta - metadata for all climatological percentiles](http://docker-dev01.pcic.uvic.ca:30402/api/multimeta?ensemble_name=upper_fraser&climatological_statistic=percentile)
* [multimeta - metadata for 5th percentiles](http://docker-dev01.pcic.uvic.ca:30402/api/multimeta?ensemble_name=upper_fraser&climatological_statistic=percentile&percentile=5)
* [multimeta - metadata for 95th percentiles](http://docker-dev01.pcic.uvic.ca:30402/api/multimeta?ensemble_name=upper_fraser&climatological_statistic=percentile&percentile=95)
* [multimeta - metadata for 75th percentile (empty set)](http://docker-dev01.pcic.uvic.ca:30402/api/multimeta?ensemble_name=upper_fraser&climatological_statistic=percentile&percentile=75)

**/data**
* [data - long term average for a mean dataset](http://docker-dev01.pcic.uvic.ca:30402/api/data?ensemble_name=ce_files&model=CNRM-CM5&variable=fdd&emission=historical,rcp85&timescale=yearly&time=0&area=)
* [data - error when requesting a percentile but not specifying which one](http://docker-dev01.pcic.uvic.ca:30402/api/data?ensemble_name=ce_files&model=CNRM-CM5&variable=fdd&emission=historical,rcp85&timescale=yearly&time=0&climatological_statistic=percentile&area=)
* data - long term average for 5th percentile
* data - long term average for 95th percentile

**/multistat**
multistats - stats for mean datasets
multistats - error when asking for percentiles, but no saying which one
multistats - stats for 5th percentile datasets
multistats - stats for 95th percentile datasets

rod-glover:
Before I review code, I'll comment on the requirements and reasoning:

> **Question for reviewers**
> Three APIs, `multimeta`, `multistat`, and `data` currently support the `climatological_percentile` attribute. These APIs  have a caller describe the dataset(s) they are interested in as a collection of filtering parameters; the other APIs return data on a single datafile and require its `unique_id`.
> 
> I added the `percentile` parameter to all three of these APIs. My reasoning was that it was good to have the APIs have similar interfaces. However, I'm not sure about this decision and would like thee reviewer to weigh in. Here are the considerations:
> 
> 1. a user can specify `climatological_statistic` as `percentile` to access data or metadata from all percentile datasets at once, if allowed. A user can additionally specify `percentile` as `5` to see only 5th percentile datasets.

:+1: 

> 
> 1. the PCEX frontend needs the `multimeta` API to be able to return a list of all percentile datasets (this is how it learns what percentiles are available). It does not currently need the `multimeta` API to be able to provide a list of all 5th percentile datasets, but I could imagine this theoretically being useful sometime.

:+1: I can imagine it being useful sometime too.

> 
> 1. The PCEX frontend needs to be able to request a specific percentile from the `data` API. The `data` API cannot display data from multiple percentiles at once (doesn't fit in its "run" format).

:+1: Good choice not to reformulate this at this time. But I think that format (by which I think you mean its response data structure) is not a good one and one fine day it would be nice to change it. Also I want a pony.

> 
> 1. The PCEX frontend does not need to be able to request either a specific percentiles or all percentiles from the `multistat` API, 

It doesn't seem impossible to me that the FE might in some future scenario request percentiles, on the principle that if it can request other statistics, why not percentiles?

> and furthermore, requesting standard deviation of a percentile dataset is kind of a mathematical mess and I don't think we'll ever want to do it.

Agreed. But I don't understand how "standard deviation of a percentile dataset" would come about, even with the new parameters. Can you explain?

> 
> Given that, which of the three APIs should I add the new `percentile` parameter to? `data` is the one we need. We could want `multimeta` someday. We'll probably never want it for `multistat` and I only put it there for consistency.

I would say `data` and `multimeta` for sure, and `multistat` pending an improved understanding of your thoughts on it.

corviday:
> Good choice not to reformulate this [data query response format] at this time. But I think that format (by which I think you mean its response data structure) is not a good one and one fine day it would be nice to change it. Also I want a pony.

Yeah, the `data` query response format is kind of limiting. An unrelated issue that gets on my nerves is that it doesn't return any metadata, so there's there sort of contortion in the front end where it has to keep the metadata describing any numbers it gets from `data` in separate, parallel structures.

> I don't understand how "standard deviation of a percentile dataset" would come about, even with the new parameters. Can you explain?

The data I am working with now is from the CanESM2-LE, the CanESM2 Large Ensemble, containing fiftyish individual runs. Hydrology has provided me with datasets representing the ensemble mean (which behaves like data we already have) and datasets representing the 5th and 95th percentiles of the model ensemble (as requested by MOTI), which is what has prompted all of these backend changes. The "percentile" datasets have the same format as regular datasets, they have timestamps and latitude and longitude and all that. So you could use the `multistat` API to determine the mode, mean, standard deviation, etc of a particular timestamp and area for these "ensemble percentile" datasets.

As of the current state of this PR, you can do that with the API. So my question is, is this _such_ an mathematically incoherent thing to do that I should have an error message when someone tries it?

rod-glover:
I like the idea of an error message. But it won't always be an error; only in the case of already-percentile datasets. If you can test for that case, and raise an error, then that sounds like a good solution to me.

ce/api/data.py
```diff
@@ -124,6 +128,21 @@ def data(
                 climatological_statistic
             )
         )
+    if percentile is not None:
+        try:
+            percentile = float(percentile)
+        except ValueError:
+            raise Exception(
+                "Percentile parameter {} not convertable to a number".format(percentile)
+            )
```
rod-glover, line 137: `Exception` is the most generic possible exception to raise. I think Python preferred practice is to be more specific, in this case maybe `ValueError`?

ce/api/multistats.py
```diff
@@ -95,6 +95,13 @@ def multistats(
             }
     """
 
+    # we don't want people requesting statistics on percentile datasets;
+    # that would be meaningless.
+    if climatological_statistic == "percentile":
+        raise Exceptions(
```
rod-glover, line 101: `Exceptions` plural?

ce/api/util.py
```diff
@@ -133,14 +133,39 @@ def check_climatological_statistic(
...
+    def final_method(p):
+        """get the name of the final cell method, corresponding 
+        to climatological aggregation (usually)"""
+        return p[-1].method.name
+
+    def final_params(p):
+        """get the parameters of the final cell method, providing details
+        of the climatological aggregation (ie, percentile value)"""
+        return p[-1].method.params
+
```
rod-glover, line 151: :+1:

ce/api/multistats.py
```diff
@@ -95,6 +95,13 @@ def multistats(
...
+            "Statistical calculations are not meaningful on percentile datasets."
+        )
+
```
rod-glover, line 104: If I understand correctly, `multistats` computes statistics as requested by the param `climatological_statistic` (which can take values of `"mean"`, `"standard_deviation"`, `"percentile"`) on the base dataset. And we are concerned to exclude the case of (any) statistic over a base dataset that is already a percentile? If so, I don't understand how this condition checks that.

ce/tests/test_util.py
```diff
@@ -296,6 +296,95 @@ def test_check_climatological_statistic(
...
+        check_climatological_statistic(
+            cell_methods,
+            climatological_statistic,
+            default_to_mean=True,
+            match_percentile=match_percentile,
+        )
+        == expected
+    )
+
+
```
rod-glover, line 387: :+1:

rod-glover (commented):
All good except my question about the error condition we most recently discussed (statistic over a percentile dataset).

ce/api/multistats.py
corviday: `climatological_statistic` is a parameter describing a dataset; it describes the aggregation that was done to make a climatology. So this parameter, along with things like `model` and `variable`, indicates which dataset(s) the caller wants statistics from.

`multistat` always returns the same set of statistics (mean, stdev, min, max, median) on whatever datasets it is working on.

This check raises an error if the user attempts to calculate those statistics on a dataset which was created as an ensemble percentile, on grounds that these statistics are meaningless and the caller doesn't want them.

rod-glover: Ah, I understand now. Sorry for the distraction. LGTM!

PR #199, opened 2022-02-23
Adding downstream api

### LINKED ISSUE #197: add an API endpoint that returns the downstream path from a selected point
Currently the "watershed" API [returns a geoJSON `Polygon` representing the boundary of all grid squares upstream from a selected point](https://services.pacificclimate.org/pcex/api/streamflow/watershed?ensemble_name=upper_fraser_watershed&station=POINT+(-121.484375+53.645834)). This polygon can be  used by the front end to display the selected watershed on a map, or to query a different API about climate data within a watershed.

We would like a new API that returns a geoJSON LineString representing the path downstream from a selected point to the sea, or to the edge of the flow map. The edge of the map or the mouth of a river into the sea will have a value of "0" in the flow map netCDF file.  This will be useful to the BC-SRIF project for showing the route the salmon take to get to the sea.

The new API should accept a `Point` in WKT format, and the name of the ensemble containing the flow map.

This downstream path will be displayed on the portal map, similar to this mockup:

gridmmap

adasungar:
Create downstream api to fix #197


ce/api/streamflow/downstream.py
```diff
@@ -0,0 +1,163 @@
...
+            return downstream_tuple
+
+        cell_routing = routing[downstream_tuple[-1]]
+        downstream_neighbour = vec_add(downstream_tuple[-1], direction_map[int(cell_routing)])
+
+        if downstream_neighbour in downstream_tuple or not is_valid_index(downstream_neighbour, routing.shape):
+            return downstream_tuple
+
+        downstream_tuple += (downstream_neighbour,)
+        return downstream(downstream_tuple)
```
rod-glover, line 161: That's very nice. I like how you've rolled `visited` into `downstream_tuple`. I also like how you are using the last element of `downstream_tuple` as the starting point of each recursion. 

My only suggestion here would be to find a more elegant name for `downstream_tuple` -- perhaps `stream`? In general, it's best to avoid putting the data type (in this case, tuple) in the name; instead just choose a name that conveys its purpose or meaning. Amongst other things, you might change the datatype later, even if its meaning is the same.

rod-glover: Aside: This is what's known as a tail recursion, and it's the recursive equivalent of a loop in procedural programming. 

You could transform this algorithm into a loop, but there's no reason to; recursion is fine too. Some compilers actually recognize tail recursion and compile to loop code, which is generally more efficient in space and time (by a constant factor, not a fundamental algorithmic difference like linear time vs quadratic time). Not sure if Python does that.

rod-glover (approved):
Great work, @adasungar ! Just a minor variable name change and then you can merge.

PR #200, opened 2022-02-23
Add Streamflow Connectivity API

### LINKED ISSUE #198: add an API endpoint that provides streamflow connectivitiy upstream of a given point
Currently the "watershed" API accepts a point in the watershed, calculates all the grid squares upstream of that point, and  [returns a geoJSON Polygon representing the outer boundary of all the grid squares upstream from a selected point](https://services.pacificclimate.org/pcex/api/streamflow/watershed?ensemble_name=upper_fraser_watershed&station=POINT+(-121.484375+53.645834)). This polygon can then be used by the front end to query a different API about climate data within a watershed.

With the salmon project, we are not just interested in average climate data across a watershed, but in how the streams in a watershed interconnect, which may help scientists understand "choke points" the salmon have to pass through on the way to upstream spawning grounds. We need a new stream network API that, similarly to the watershed API, takes a point and returns data about the squares "upstream" of that point, but instead of the outline of the entire collection of squares, it needs to return a graph description of the squares showing how they are interconnected by the movement of water.

The new API should accept a` Point `in WKT format, and the name of the ensemble containing the flow map.

~~There are many possibilities for describing the flow network graph in JSON format. One possibility is:~~

```JSON
{
  "graph": {
    "nodes": [
      {
        "latitude": 50,
        "longitude": 123,
      },
      {
        "latitude": 51,
        "longitude": 123
      },
    ],
    "edges": [
      [1, 0],
    ]
  }
}
```

~~Which would mean there were two grid cells in this watershed, one at (50, 123) and one at (51, 123) and that the more northern one (node 1)  flowed into the more southern one (node 0). The format does not have to be identical to this - there are lots of options - but it should be straightforward enough for the front end to easily draw the network from whatever JSON it receives from this API.~~ 

**EDIT**: return GeoJSON instead, as suggested by @jameshiebert below.

 This output will be used by the web frontend to produce a map that looks something like this:

flowmap(1)

This API will do a lot of the same calculations as the watershed API - both of them trace through the flow network netCDF to find all upstream points, before returning information on those points. It will probably make sense to move some functions from the watershed API into a stream flow library file that both APIs can draw from.

helfy18:
Add an API function that shows how individual streams that are upstream flow into a cell. Resolves #198


Pipfile
```diff
@@ -23,7 +23,7 @@ cf-cell-methods = "==0.1.0"
 Sphinx = "==3.2.1"
 m2r2 = "==0.2.5"
 modelmeta = "*"
-ce = "*"
+ce = {editable = true, path = "."}
```
rod-glover, line 26: Suggest this go in `[dev-packages]`. Logically, it belongs there because it is required for development, and is not a dependency for installing the package itself (i.e., a package is not a dependency of itself).

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,217 @@
...
+        sesh, ensemble_name, "flow_direction"
+    ) as flow_direction_ds, get_time_invariant_variable_dataset(
+        sesh, ensemble_name, "elev"
+    ) as elevation_ds, get_time_invariant_variable_dataset(
+        sesh, ensemble_name, "elevmin"
+    ) as elevation_min_ds, get_time_invariant_variable_dataset(
+        sesh, ensemble_name, "elevmax"
+    ) as elevation_max_ds, get_time_invariant_variable_dataset(
+        sesh, ensemble_name, "area"
+    ) as area_ds:
```
rod-glover, line 91: Every so often Black doesn't produce very elegant layout. :stuck_out_tongue_winking_eye:

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,217 @@
...
+
+    lines = MultiLineString(watershed_lonlats)
+
+    return {
+        "streams": geojson_feature(lines,),
+        "debug/test": {"watershed_streams": {"time": watershed_time.elapsed,}},
+    }
+
+
+def build_watershed_streams(target, routing, direction_map, debug=False):
```
rod-glover, line 158: As we discussed on Zulip, this function is very similar to the watershed code. Can you abstract out the repeated/common part and share it between them? That would make any fixes apply to both, and reduce or eliminate duplication of testing code.

rod-glover: If there are other commonalities, say in the worker or other higher level code, it _may_ be worth factoring that out too.

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,171 @@
...
+                neighbour, cell, routing, direction_map
+            ):
+                upstream(neighbour)
+                connection.append([])
+                i += 1
+                connection[i].append(cell)
+        return connection
+
+    connection = upstream(target)
+    return set(frozenset(i) for i in connection if len(i) > 1)
```
rod-glover, line 171: @helfy18 , kindly have a look at the [discussion in the issue](https://github.com/pacificclimate/climate-explorer-backend/issues/198#issuecomment-1054706359) regarding what linestrings should be returned. I think this code solves a more difficult and unnecessary problem. Could you revise it to solve the simpler problem?

rod-glover (changes requested):
The factoring out of the common code looks good. The problem to be solved has been simplified; see my comment below about that.

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,171 @@
...
+    - Routing graphs can and in practice do contain cycles. Variable `visited`
+    is used to determine whether a cell has already been visited during the
+    traversal of the routing graph, i.e., whether we are cycling, and if so
+    not to repeat that subgraph.
+    """
+    visited = set()
+    i = 0
+    connection = [[]]
+
+    def upstream(cell):
```
rod-glover, line 151: I'm assuming that we are continuing with the "long streams" approach as opposed to the "length-2 line segments" approach discussed in the issue.

I'm having a difficult time deciding if this function does what (I think) it is supposed to do. It's not at all obvious to me whether the recursive calls to `upstream` leave `i` and `connection` in correct states in all cases. Can you explain how that works?

rod-glover (changes requested):
Johnathan, a couple of minor items, plus a question about the core algorithm.

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,171 @@
...
+            abort(
+                404,
+                description="Station lon-lat coordinates are not within the area "
+                "for which we have data.",
+            )
+
+
+def worker(
+    station_lonlat, flow_direction,
+):
```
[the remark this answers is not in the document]
helfy18: On this repository Black is always used as a pre-commit action, so it was definitely run, but we can try and fix the setup

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,171 @@
...
+        connection[i].append(cell)
+        for neighbour in neighbours(cell):
+            if neighbour not in visited and is_upstream(
+                neighbour, cell, routing, direction_map
+            ):
+                upstream(neighbour)
+                connection.append([])
+                i += 1
+                connection[i].append(cell)
+        return connection
```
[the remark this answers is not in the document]
helfy18: When I removed this ```return connection``` I began getting an error that ```NoneType is not iterable```.

ce/api/streamflow/watershed_streams.py
helfy18: When a cell has an eligible neighbor, a recursive call is made on the eligible neighbor. This will cause all of the cells on one path to be put into a list. When ```connection.append([])``` happens, this means that there are no longer any eligible upstream cells (since we will have returned from the consistent calls of ```upstream(neighbour)```). We then start a new path from the furthest point upstream that has a neighbor. This cell is added to the path, and we continue the process, until it returns it's way back down to the initial cell. After the ```upstream``` is done running, connection contains a list of lists, the inner lists being the path's of cells. I made a mistake overloading ```i```. I have updated it so that the variable used in ```upstream``` is called counter, as this variable counts the number of paths in ```connection``` (so it knows where to put the next cell that is upstream). The ```i``` that was used in the return statement has been renamed to ```stream```, as this is a more fitting name.

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,164 @@
...
+        watershed_xys = build_watershed_streams(
+            flow_direction.lonlat_to_xy(station_lonlat),
+            flow_direction.values,
+            direction_matrix,
+            debug=True,
+        )
+
+    # `watershed_lonlats` must be an ordered collection (not sets) because
+    # a multi line string is an array (python list) of linestrings
+    watershed_lonlats = [[]]
```
rod-glover, line 100: We no longer need this line; it's a leftover from the previous way of computing `watershed_lonlats` .

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,164 @@
...
+                neighbour, cell, routing, direction_map
+            ):
+                upstream(neighbour)
+                connection.append([])
+                counter += 1
+                connection[counter].append(cell)
+        return connection
+
+    connection = upstream(target)
+    return set(frozenset(stream) for stream in connection if len(stream) > 1)
```
rod-glover, line 164: Why are these turned into frozensets? And not left as just lists or transformed into tuples, the immutable version of lists?

rod-glover (changes requested):
Again, a few changes plus a question about the upstream algorithm.

ce/api/streamflow/watershed_streams.py
helfy18: this was done for ease of testing, so that the order didn't matter when writing tests. It would probably make sense to just keep it as list of lists the entire time and write the tests to be exact, as they are for the ```worker``` class. I can switch these if that is what you think is best

ce/api/streamflow/watershed_streams.py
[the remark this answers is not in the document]
helfy18: that is correct

ce/api/streamflow/watershed_streams.py
helfy18: that is correct

rod-glover: We apparently have an algorithm that works (tests pass :tada:), and we have an explanation of how it works.

That it works is good. Now we have to ask how readable and maintainable it is, because it is part of a living codebase. My thoughts on this, in descending order of importance:

1. It has what's often called a "code smell", which in this case is the creation of unused objects (`connection`s) which have to be cleaned up at the end. Creating things that have no use, even transiently, suggest a tighter algorithm can be developed.
2. Add comments that explain how the algorithm works. It's not instantly obvious. I suggest a sort of global explanation akin to the ones you and I formulated, plus the occasional comment on a key line in the body of the code.
3. Some algorithms need explanations because they are inherently tricky, or rather because the problem they solve forces some trickiness. This may be one such. But at this point, we should wonder if there is an algorithm whose operation is clearer and which needs less explanation. The benefit being to later maintainers who need to modify what it does. Which leads into the next observation: 
4. This is an example of a recursive algorithm with global (nonlocal) variables that it _modifies_ -- they are the result it generates. 
   - That kind of algorithm is notoriously tricky to write and understand: evidenced by the fact that I had to spend a few minutes thinking about how it works and if it works in all cases and to confirm that understanding with you. 
   - Recursive algorithms are more "natural" and comprehensible when they don't use global variables and instead restrict themselves to values they receive in arguments. These are called purely functional (no side effects; no global variables). They construct their results from only their arguments and the return values of recursive calls. You only have to establish that the algorithm works on the local information not on hard to establish guarantees that the global variables are always in a suitable state at each recursive call. 
   - There is (by first principles) an equivalent purely functional algorithm. Whether such an algorithm would actually be simpler to write and understand is an open question, but it's worth giving it some thought. An exercise for the student, so to speak.
5. Finally, a bit around naming: Leaving aside the question of whether we go to a purely functional algorithm, we have variables named `connection` and `counter`. But when you explain the algorithm, you use the terms "streams" and "number of streams". It's always better for the explanation and the code to be congruent, i.e., to use the same terminology.

What I suggest at this point is the following:
- Set a deadline of 1 week for any renovations to this algorithm plus the other few outstanding comments/changes. If there are still questions, problems, or tests not passing, then we go back to the current algorithm.
- Fix items 1 and 2 above.
- Think about items 3 and 4, and at least outline your thinking to me. This is intended to be a learning experience -- both algorithmically and about what it means to participate in a long-lived codebase. If we agree it's worthwhile, we can update the code accordingly.
- Do 5 as appropriate to the algorithm finally arrived at.

rod-glover (changes requested):
Better. And, some further comments about the `upstream` algorithm. 

Also let's get the other outstanding comments/questions from me cleared up too. Either put in a reply or change the code. I can resolve the convo's that don't result in a code change.

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,194 @@
...
+        # be added to the current stream.
+        if eligible == [] and first:
+            return
+        elif first:
+            streams.append([cell])
+        else:
+            streams[number_of_streams].append(cell)
+
+        # At this point, the first cell is finished, so the flag is no longer
+        # true
```
rod-glover, line 183: Delete this comment.

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,194 @@
...
+    streams = []
+    number_of_streams = 0
+
+    def upstream(cell, first):
+        """Generates list of lists, each list being a stream of cells upstream of
+        the inputted cell. This list is named 'streams', a nonlocal variable.
+
+        - Variable 'eligible' is generated for each cell, and is a list of all
+        'eligible' neighbours. Eligible is defined as a neighbour that is
+        upstream of the current cell and that also has not been visited.
```
rod-glover, line 157: I think this comment would be better placed right where eligible is computed.

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,194 @@
...
+    This variable is used to ensure the cells are inserted into the right stream
+    """
+    visited = set()
+    streams = []
+    number_of_streams = 0
+
+    def upstream(cell, first):
+        """Generates list of lists, each list being a stream of cells upstream of
+        the inputted cell. This list is named 'streams', a nonlocal variable.
+
```
rod-glover, line 154: Document the meaning of `first` here.

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,194 @@
...
+            neighbour
+            for neighbour in neighbours(cell)
+            if neighbour not in visited
+            and is_upstream(neighbour, cell, routing, direction_map)
+        ]
+
+        # If there are no eligible neighbours for the first cell, connection
+        # should be []. If there are eligible neighbours, the first stream
+        # must be started. Otherwise, the cell is a part of a stream and can
+        # be added to the current stream.
```
rod-glover, line 174: Good explanation.

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,194 @@
...
+            streams.append([cell])
+        else:
+            streams[number_of_streams].append(cell)
+
+        # At this point, the first cell is finished, so the flag is no longer
+        # true
+
+        # For each eligible neighbour, we want to add them to a stream and
+        # move upstream on those neighbours (call the function again)
+        for neighbour in eligible:
```
rod-glover, line 187: Add a comment that explains that the first neighbour gets added to the current stream, and that all other neighbours start new streams. That's the crux of this algorithm, and saving a reader from having to figure that out is gold.

ce/api/streamflow/watershed_streams.py
```diff
@@ -0,0 +1,194 @@
...
+            upstream(neighbour, False)
+            if neighbour != eligible[-1]:
```
rod-glover, line 189: Add a little comment explaining why you don't want to start a new stream after the last item.

rod-glover (approved):
Great work, @helfy18 ! It's been a long haul on this PR!

Once you have addressed the changes I requested (documenting the code a little more), you can merge.

PR #213, opened 2022-05-31
downgrade rasterio

corviday:
Upgrading `rasterio` to version 1.2.6 seems to have broken some functionality - specifically this version of rasterio cannot read the new Fraser watershed file for unknown reasons. This downgrade restores the previously pinned version of rasterio.

Resolves https://github.com/pacificclimate/climate-explorer-backend/issues/209


corviday:
This has not fixed the issue. Further investigation needed.

PR #223, opened 2023-01-30
Add pre-ping param to engine creation

### LINKED ISSUE #220: Try connection pre-ping
We are having trouble with connections being dropped by our proxy (or some other culprit), leading to an error in the backend. [SQLAlchemy connection pre-ping](https://docs.sqlalchemy.org/en/20/core/pooling.html#disconnect-handling-pessimistic) is one possible "external" answer to that problem. However, note the following from that document:

> It is critical to note that the pre-ping approach does not accommodate for connections dropped in the middle of transactions or other SQL operations. If the database becomes unavailable while a transaction is in progress, the transaction will be lost and the database error will be raised. While the [Connection](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Connection) object will detect a “disconnect” situation and recycle the connection as well as invalidate the rest of the connection pool when this condition occurs, the individual operation where the exception was raised will be lost, and it’s up to the application to either abandon the operation, or retry the whole transaction again.

There is an alternative; see #221 .

rod-glover:
Resolves #220 

Testing seems to show that this fixes the 500 problem observed for the last few months.

### DIFF
--- a/ce/__init__.py
+++ b/ce/__init__.py
@@ -13,6 +13,7 @@ def get_app():
     app.config["SQLALCHEMY_DATABASE_URI"] = os.getenv(
         "MDDB_DSN", "postgresql://httpd_meta@monsoon.pcic.uvic.ca/pcic_meta"
     )
+    app.config["SQLALCHEMY_ENGINE_OPTIONS"] = {"pool_pre_ping": True}
     app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
     add_routes(app)
     return app


corviday:
Merge after tonight's planned tests?

rod-glover:
Yes, waiting for that second confirmation. Still in Sherlock mode: everyone is still a suspect.

corviday (approved):
LGTM!

PR #226, opened 2023-09-14
I225 restructure query joins

### LINKED ISSUE #225: Query joins need restructuring
I ran into a query-related issue within ce/api/utils.py, which is used in multistats.py
The error was:
`SQLAlchemy Error: Query.join() takes from 2 to 3 positional arguments but 9 were given`

[This query](https://github.com/pacificclimate/climate-explorer-backend/blob/8247c55c73d516a492d426c96f09b769f88f024c/ce/api/util.py#L254C5-L273C6) was the source of my error, and I fixed it by splitting the large join into a chained version like this:

```
query = (
        sesh.query(mm.DataFile.unique_id)
        .distinct(mm.DataFile.unique_id) 
        .join(mm.DataFileVariableGridded)
       ⋮
    ) 
```
Queries should be restructured as above and in accordance with the [docs](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.join).

QSparks:
This fixes two incorrect queries in the Climate Explorer utilities and changing the version of `pipenv` resolves python-CI issues during dependency installation.

Resolves #225

### DIFF
--- a/ce/api/util.py
+++ b/ce/api/util.py
@@ -34,13 +34,11 @@ def get_units_from_run_object(sesh, run, varname, ensemble_name):
     units = (
         sesh.query(mm.VariableAlias.units)
         .distinct(mm.VariableAlias.units)
-        .join(
-            mm.DataFileVariableGridded,
-            mm.EnsembleDataFileVariables,
-            mm.Ensemble,
-            mm.DataFile,
-            mm.Run,
-        )
+        .join(mm.DataFileVariableGridded)
+        .join(mm.EnsembleDataFileVariables)
+        .join(mm.Ensemble)
+        .join(mm.DataFile)
+        .join(mm.Run)
         .filter(mm.Ensemble.name == ensemble_name)
         .filter(mm.DataFileVariableGridded.netcdf_variable_name == varname)
         .filter(mm.Run.name == run.name)
@@ -79,7 +77,6 @@ def open_nc(resource):
 
 
 def get_array(nc, resource, time, area, variable):
-
     if variable not in nc.variables:
         raise Exception(f"Resource {resource} does not have variable {variable}.")
 
@@ -140,7 +137,7 @@ def check_climatological_statistic(
     """
 
     def final_method(p):
-        """get the name of the final cell method, corresponding 
+        """get the name of the final cell method, corresponding
         to climatological aggregation (usually)"""
         return p[-1].method.name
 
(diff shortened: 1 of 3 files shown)


ce/api/util.py
```diff
@@ -248,22 +245,21 @@ def search_for_unique_ids(
     )
 
     matching_cell_methods = filter_by_climatological_statistic(
-        [r[0] for r in cell_methods], climatological_statistic,
+        [r[0] for r in cell_methods],
+        climatological_statistic,
```
rod-glover, line 249: Aside: Black respects a trailing comma by placing each item on a separate line, regardless of whether they would all fit on one line. You may or may not want it like this. No big deal.

ce/api/util.py
```diff
@@ -248,22 +245,21 @@ def search_for_unique_ids(
...
-            mm.Time,
-        )
+        .join(mm.DataFileVariableGridded)
+        .join(mm.EnsembleDataFileVariables)
+        .join(mm.Ensemble)
+        .join(mm.Run)
+        .join(mm.Model)
+        .join(mm.Emission)
+        .join(mm.TimeSet)
+        .join(mm.Time)
```
rod-glover, line 262: Good. Are there one or more tests that exerise this code? I'd recommend that there is.

rod-glover (approved):
LGTM! All that Pipenv ugliness. 

Only caveat would be if there aren't test(s) for the changed code.

ce/api/util.py
QSparks: Thanks, Rod! Tests are exercising both sections of modified code. The  [first query](https://github.com/pacificclimate/climate-explorer-backend/blob/307a8a87d47049e00042735a4344c52d6c276249/ce/api/util.py#L35C9-L35C10) is tested in [test_utils](https://github.com/pacificclimate/climate-explorer-backend/blob/8247c55c73d516a492d426c96f09b769f88f024c/ce/tests/test_util.py#L127) while the [second query](https://github.com/pacificclimate/climate-explorer-backend/blob/307a8a87d47049e00042735a4344c52d6c276249/ce/api/util.py#L252) is checked during [test_multistats](https://github.com/pacificclimate/climate-explorer-backend/blob/8247c55c73d516a492d426c96f09b769f88f024c/ce/tests/api/test_multistats.py#L45C6-L45C6) and is called from [multistats.py](https://github.com/pacificclimate/climate-explorer-backend/blob/8247c55c73d516a492d426c96f09b769f88f024c/ce/api/multistats.py#L105).

PR #229, opened 2023-11-07
Accept POST requests

### LINKED ISSUE #227: accept area polygons in body of requests
Applications that allow users to select regions of interest by clicking on a point and selecting all grid squares upstream of that point (PCEX extreme streamflow portal, SCIP front end) wish to request data on polygons with a lot of vertices. Putting the polygon in the URL as the `area` parameter can result in URLs longer than the informal 4096 limit.

These front ends should instead put their long polygons in the `body` of the request, and put a hash of the polygon string in the url as the parameter `area-hash` (to support caching). This backend should determine the requested area as follows:

1. If the `area` parameter is specified, use that polygon
2. If the `area-hash` parameter is specified and there is a polygon in the `body` of the request, use that polygon
3. otherwise, process the request across the entire domain

corviday:
Updates the backend to accept HTTP `POST` requests in addition to HTTP `GET`. This is a straightforward update thanks to Flask's handy `request.values` attribute, which contains all parameters supplied either in the URL of a `POST` or the body of a `GET`, allowing the request methods to be treated identically.

The goal here is to be able to accept longer `area` strings, especially from user-selected regions, such as when a user clicks on a point in a flow network to select all grid squares upstream of that point. One notable example of selecting an upstream area along the FRASER river results in an `area` string ten thousand characters long, too long to send in a URL.

resolves #227

### DIFF
--- a/ce/api/__init__.py
+++ b/ce/api/__init__.py
@@ -50,22 +50,22 @@ __all__ = list(methods.keys()) + ["call"]
 
 def call(session, request_type, item=None):
     """Extracts request query parameters, checks for required arguments
-       and delegates to helper functions to fetch the results from
-       storage
+    and delegates to helper functions to fetch the results from
+    storage
 
-       Args:
-          session (sqlalchemy.orm.session.Session): A database Session object
-          request_type(str): name of the API endpoint to call
-          item(str): name of an individual item for a REST API
+    Args:
+       session (sqlalchemy.orm.session.Session): A database Session object
+       request_type(str): name of the API endpoint to call
+       item(str): name of an individual item for a REST API
 
-       Returns:
-          werkzeug.wrappers.Response.  A JSON encoded response object
+    Returns:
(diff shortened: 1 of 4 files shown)


corviday:
A demo is available at https://services.pacificclimate.org/dev/pcex/api/

Sample curl commands for testing:
```
# a POST request
curl -X POST 127.0.0.1:8000/api/timeseries -F 'id_=cdd_sClimMean_BCCAQv2_PCIC12_historical-rcp85_rXi1p1_19610101-19901231_Canada' -F 'area=""' -F 'variable=cdd'

# identical GET request
curl "http://127.0.0.1:8000/api/timeseries?id_=cdd_sClimMean_BCCAQv2_PCIC12_historical-rcp85_rXi1p1_19610101-19901231_Canada&area=&variable=cdd"
```

jameshiebert (approved):
Looks pretty straightforward to me. Nice work!

PR #230, opened 2023-11-17
Convert dependency management to Poetry

jameshiebert:
This PR removes the Pipenv config and adds the necessary config to use Poetry for dependency management under Python 3.8 (we can add more modern versions of Python as a follow up).

A few things to note:

## Dependencies

The dependencies are getting a little elderly. I did *as little upgrading as possible* to make everything work, but once we're happy with the Poetry setup, we should prioritize the elimination of deprecation warnings and perform upgrades to the full dependency chain in the near future.

I had to make some very minor changes to a couple uses of `flask` and `werkzeug` to accommodate API changes in the minimum installable versions. Very minor though.

## Testing

Apparently :frowning_face: the `climate-explorer-backend` has acquired some online tests that require live data and a database connection. [All of the offline tests pass](https://github.com/pacificclimate/climate-explorer-backend/actions/runs/6907256137/job/18794017518). To test the online tests:

The live data (`REGIONS_DATA_DIRECTORY`) are relatively small and can be copied to a workstation for testing (`REGIONS_DATA_DIRECTORY=/path/to/data pytest -v -m "not online"`)

If one has database access and access to filesystem data, the full testing is something like this:
`REGION_DATA_DIRECTORY=/storage/data/projects/comp_support/climate_explorer_data_prep/precalculated_regions/ MDDB_DSN=postgresql://ce_meta_ro@db.pcic/ce_meta poetry run py.test -v`

The Docker file `docker/Dockerfile.dev` should work for testing.

```
docker build -t ce-backend:poetry -f docker/Dockerfile.dev .
docker run -it --rm --name ce-backend-testing --mount type=bind,source=/storage/data/climate,target=/storage/data/climate -v /storage/data/projects:/storage/data/projects:rw ce-backend:poetry poetry run /bin/sh
```


docker/Dockerfile.dev
```diff
@@ -1,18 +1,21 @@
...
 WORKDIR /app
-RUN pip3 install -U pip pipenv --ignore-installed
-RUN pipenv install
 
-ADD . /app/
+RUN apk add libffi-dev
+RUN wget -O - https://install.python-poetry.org | python3 -
+ENV PATH=/root/.local/bin:$PATH
+RUN poetry install --with=dev
+RUN poetry install
```
rod-glover, line 12: Is this still necessary? What was the outcome of `install` vs `install --with`? The documentation sure looks as if it shouldn't be necessary. Also, the CI workflow doesn't use it. :thinking:

pyproject.toml
```diff
@@ -0,0 +1,41 @@
+[tool.poetry]
+name = "ce"
+version = "3.6.1"
+description = "PCIC's Climate Explorer (CE)"
+authors = ["James Hiebert <hiebert@uvic.ca>"]
```
rod-glover, line 5: Should we add others?

rod-glover (approved):
Looks great. Nice and simple, and evidently it works.

corviday:
Maybe update README.md with new installation instructions?

docker/Dockerfile.dev
jameshiebert: Nope. Wasn't necessary after all. Thanks for checking.

pyproject.toml
jameshiebert: Good suggestion! Yes, that metadata hadn't been updated since the first commit years ago.

jameshiebert:
> Maybe update README.md with new installation instructions?

Good suggestion, @corviday! Done.

Issue #233, opened 2024-02-20
sqlalchemy error


corviday:
Docker instances of the newest version (3.6.2) throw the following error when any API call is made:

```
2024-02-20 17:03:47 [13] [ERROR] Exception on /api/multimeta [GET]
Traceback (most recent call last):
File "/root/.cache/pypoetry/virtualenvs/ce-9TtSrW0h-py3.8/lib/python3.8/site-packages/sqlalchemy/util/_collections.py", line 1020, in __call__
return self.registry[key]
KeyError: 140433640835912
```

This does not seem to be an issue with our python code, as a version with identical code has been deployed as an unmerged branch. It is perhaps caused by an update to sqlalchemy.

corviday:
The version of sqlalchemy used by the app has not changed.

The error does not occur outside of docker

The file referenced in the error does exist

corviday:
I built a docker from `32ce883`, and it did not have the error, so I think the error was introduced sometime after that hash, though it is still unclear what sort of error might appear only in docker containers and not on bare metal.

jameshiebert:
@corviday have you tried running the test suite in a docker container? Does the `/multimeta` API call fail under that environment? Just trying to get a minimal test case so that we can narrow this down and work on it.

corviday:
That's a great idea, thanks. The test suite has some failures, but they seem to be unrelated (a needed directory isn't mounted to the container) but other tests pass. I don't know what that means, but it's useful.

jameshiebert:
OK, that being the case can you give a full "steps to reproduce" overview then?

corviday:
Steps to reproduce:

1. Build a docker of release version 3.6.2, or use the one on docker hub
2. Start the docker
3. access any API endpoint that uses the database to get an error. /multimeta is an easy one because it has defaults for all the parameters.

Given that tests work, issue is likely flask-related.

jameshiebert:
Can you be more specific about what parameters you're using for step 2 and exactly what URL you're using from step 3? I still cannot reproduce this.

corviday:
That's good news, I think?

Here are the current docker-compose.yaml and be.env files, with their extension changed and passwords redacted.
[be.env.txt](https://github.com/pacificclimate/climate-explorer-backend/files/14390881/be.env.txt)
[docker-compose.yaml.txt](https://github.com/pacificclimate/climate-explorer-backend/files/14390882/docker-compose.yaml.txt)

[...]

Issue #243, opened 2024-09-12
numpy build error


corviday:
The poetry build on github has stopped working. This is a new-ish error, first cropping up last week.

Build errors that seem relevant:

> Failed to install setuptools>=41.2, numpy>=1.26.0b1; python_version>='3.12.0.rc1', oldest-supported-numpy; python_version<'3.12.0.rc1', wheel, cython>=0.29.20.

>  Note: This error originates from the build backend, and is likely not a problem with poetry but with numpy (1.17.3) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "numpy (==1.17.3)"'.

> Cannot install build-system.requires for cftime.

`numpy 1.17.3` is import by `oldest-supported-numpy`
`oldest-supported-numpy` is imported by `cftime`
`cftime` is imported by `netcdf4`
`netcdf4` is imported by `nchelpers`
`nchelpers` is imported by `climate-explorer-backend`. 

poetry can't install `numpy 1.17.3`.

Questions:
1. How can this be fixed?
2. Why did it break Friday September 6th. when it was previously working? What changed? Is there a package we can pin to an old version to get it to work again, or did some sort of cache expire in a way we can't easily turn back the clock on?

corviday:
Things that haven't worked:

* pinning netcdf4 to 1.5*, 1.6*, or 1.7*
* explicitly pinning cftime
* explicitly pinning numpy

Things that have worked but that we don't want to do:
* downgrading from poetry to pip

corviday:
Weirdly, `poetry show --tree` doesn't show this issue at all:

```
climate-explorer-backend$ poetry show --tree | grep numpy
│   │   │   └── numpy >1.13.3 
│   │   └── numpy * (circular dependency aborted here)
│   │   └── numpy >1.13.3 
│   └── numpy * (circular dependency aborted here)
├── numpy *
│   │   └── numpy >1.13.3 
│   └── numpy * (circular dependency aborted here)
numpy 1.24.4 Fundamental package for array computing in Python
├── numpy *
    ├── numpy * 
└── numpy >=1.14
```

No mention of `oldest-supported-numpy`.

jameshiebert:
Can you add the output of `poetry env info`? Wondering what version of Python you're using.

corviday:
```
RUN poetry env info

Virtualenv
Python:         3.8.15
Implementation: CPython
Path:           NA
Executable:     NA

Base
[...]

PR #246, opened 2024-09-19
test branch for build changes #4

corviday:
This branch is using an experimental [version](https://github.com/pacificclimate/docker-geospatial-python/pull/14) of the geospatial-python container.

Changes in that container:
* ubuntu 22.04
* gdal 3.4.0
* python 3.10

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: '3.8'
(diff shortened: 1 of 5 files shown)


corviday:
The code has been updated to python 3.10, and now builds and passes tests in github containers. This hopefully puts #243 to bed. 

I am going to try deploying this branch to portainer and see if the updates to python 3.10, ubuntu 22.04, and gdal 3.4 fix #233 as well. There's no real reason to think they would, but sometimes just modernizing a stack is magic.

corviday:
This did indeed fix #233.

PR #254, opened 2026-05-13
Maintenance sprint 26q1

QSparks:


.github/workflows/python-ci.yml
```diff
@@ -2,34 +2,37 @@ name: Python CI
...
-        apt-get install -yq libxml2-dev libxslt-dev libffi-dev wget
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - name: Install system dependencies
+        run: |
+          apt-get update 
+          apt-get install -yq libxml2-dev libxslt-dev libffi-dev wget
+
+      - name: Install poetry
```
Nospamas, line 23: I don't love how we do the pipe to python thing here. Is it worth reviewing https://python-poetry.org/docs/#ci-recommendations and using pipx to install at specific versions and avoid potential security issues?

Nospamas (approved):
Good updates. Discussed briefly with Q that this is a decent first cut of a cache implementation, but longer term Redis or one of its clones might solves some of the cross process locking / coherency and improve async write behaviour.

Nospamas (approved):
Good changes, I'll keep this one in mind as I work on other backends.

PR #255, opened 2026-06-05
Add devcontainer, update to 3.14 python compatability

Nospamas:
- Add devcontainer setup (Dockerfile, config, post-install scripts)
- Fix rasterio_thredds_helper to use vsicurl and convert dodsC to fileServer URLs
- Bump rasterio dependency to >=1.4
- Refactor open_nc to use context manager instead of try/finally
- Expand CI matrix to Python 3.11–3.14 with timeout and concurrency settings
- Tweak tests to work with simplified Thredds server urls


pyproject.toml
```diff
@@ -31,13 +31,12 @@ keywords = [
 ]
 dependencies = [
     "Flask>=3.1.1,<4",
-
     "Flask-SQLAlchemy>=3.0.5,<4",
     "Flask-Cors",
```
QSparks: Should we pin?  `4.0.0`

ce/api/geo.py
```diff
@@ -263,10 +263,11 @@ def polygon_to_masked_array_helper(raster):
...
     """
-    Opens a remote NetCDF resource for Rasterio, handling OpenDAP (dodsC) or fileServer as needed.
-    Assumes Rasterio has GDAL with OpenDAP support.
+    Opens a remote NetCDF resource for Rasterio via GDAL's vsicurl.
+    THREDDS dodsC (OPeNDAP) URLs are converted to fileServer URLs since
+    GDAL's NetCDF driver requires direct HTTP file access, not OPeNDAP.
     """
-    # Construct proper NetCDF subdataset path
-    dataset_url = f'NETCDF:"{resource}":{variable}'
+    http_resource = resource.replace("/dodsC/", "/fileServer/", 1)
```
QSparks: I don't think we want a replace here.
The comment suggests it should should probably accept either a dodsC URL or a fileServer URL.

Maybe one of our tests could use the fileServer URL?

ce/tests/api/test_geospatial.py
```diff
@@ -80,30 +94,33 @@ def test_geojson_feature(thing, kwargs, expected):
...
+                    ),
+                ]
             ),
         ),
     ),
 )
 def test_outline(centres, height, width, expected):
     ol = outline_cell_rect(centres, height, width)
-    assert ol.equals(Polygon(expected))
+    assert ol.equals(expected)
```
QSparks, line 124: Should we also assert the geom_type now that it could be a Polygon or MultiPolygon?

ce/tests/api/test_geospatial.py
QSparks: OOS: This could be isinstance()

ce/tests/conftest.py
```diff
@@ -30,6 +30,12 @@
 
 from ce import get_app
 
+# Filesystem root where /testdata/ database paths are stored locally.
+# Override with TESTDATA_PATH env var if files live elsewhere.
+_TESTDATA_LOCAL = os.getenv(
+    "TESTDATA_PATH", "/storage/data/projects/comp_support/daccs/test-data"
```
QSparks: Should the fallback be the local directory: `ce/tests/data/`

QSparks (approved):
LGTM, few comments around paths, Thredds and the geo tests.

PR #256, opened 2026-06-06
Pgbouncer config healthcheck

Nospamas:
follows from #255. 
- Allows customisation of the Sqlalchemy connection pool for use with PGBouncer 
- Adds a healthcheck route with optional db check.


QSparks (approved):
LGTM!

PR #257, opened 2026-06-25
container updates, use updated gdal container, split build, updated devcontainer

Nospamas:
- Updates CE backend to use updated base container with gdal 3.12.2
- Updates to use split build (build stage/runtime stage)
- Updates to run tests in geospatial container
- Several devcontainer updates to also work with the updated container
- No longer runs as root (now 10001)


QSparks (approved):
LGTM! Very much looking forward to the new devcontainer. 😀

With the new non-root user for the prod image, will we hit write permission issues with the cache?
 `- /.../ce-multimeta-cache:/cache:rw`

QSparks:
> With the new non-root user for the prod image, will we hit write permission issues with the cache? `- /.../ce-multimeta-cache:/cache:rw`

We do run into permission issues here.

